Issue #121, opened 2014-09-11
Add package/<name>.rss to get notifications on package updates


DavidSouther:
I would love to have RSS feeds per package, that track the package and its dependencies for new versions pushed to NPM.

It looks like that could go in the registry facet, as `show-rss.js`. I'll try to take a stab at this over the weekend.

zeke:
Closing this, as the action is happening here: https://github.com/npm/newww/pull/122

PR #122, opened 2014-09-12
First pass at basic /registry/<package>.rss route.

DavidSouther:
Solution for #121 

This is a first pass, mostly to try playing with newww. It's a feature I think would be really useful. Not sure long term support or issues with it, but here's a first blush.

The registry facet gets a new route, `/registry/{package}.rss`, that returns an rss xml with reach release as an `<item>`. Long term, I'd like to see options for "deep" notifications, generating items for each dependency update as well. This way, you could subscribe to the "master" rss feed for your package, and know when your deps are updated to run tests or re-release or whatever.

### DIFF
--- a/facets/registry/index.js
+++ b/facets/registry/index.js
@@ -1,0 +1,88 @@
+var path = require('path');
+var internals = {};
+
+exports.register = function Regsitry (facet, options, next) {
+
+  facet.views({
+    engines: { hbs: require('handlebars') },
+    path: path.resolve(__dirname, 'templates')
+  });
+
+  facet.route({
+    path: "/package/{package}/{version?}",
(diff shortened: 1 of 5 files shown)


zeke:
Hey @DavidSouther, now that newww has shipped and is stabilizing, it would be awesome to see this feature land on the site. Any interesting in bringing this PR up to date?

DavidSouther:
Absolutely! I've got some time today and this weekend - if you don't hear from my by Monday or Tuesday, want to ping me then?

zeke:
Sounds good. :)

zeke:
@DavidSouther I'm spring cleaning this PR. Feel free to update and re-open at a later time.

Issue #233, opened 2014-11-10
fix deployment with shrinkwrap or similar


rockbot:
node_modules isn't getting updated like we want it to, so we need to force it to

PR #236, opened 2014-11-11
Shrinkwrap

rockbot:
Addresses #233 

While it's technically ready for merging, I'd like to discuss how to make it a regular part of deployment instead of having to remember to remove/remake npm-shrinkwrap.json every time. (Because humans are forgetful creatures :-P)


zeke:
Gonna close this. We can revisit later if needed.

PR #293, opened 2014-12-01
use npm-collection-* packages for homepage sections

zeke:
This is an experiment in using npm packages as collections. It is an interim solution to the "we can't fetch metadata for a bunch of packages at once at runtime" problem. 

The first two collections:

https://github.com/npm/npm-collection-staff-picks#readme
https://github.com/zeke/npm-collection-explicit-installs#readme

It is the responsibility of each collection to fetch its own registry metadata before being published. Enter [backfill](https://www.npmjs.org/package/backfill), a thing that looks for all `dependencies` in `package.json` and downloads registry metadata to a file called `meta.json`. Each collection has a `postinstall` script that writes `meta.json` automatically, and sets the `meta.json` file as its `main`. The result of this rigamarole is self-contained collection packages and simple/DRY inclusion of these collections in the newww codebase...


zeke:
This approach doesn't scale. I'm writing it off as a failed experiment.

Issue #78, opened 2014-08-22
consider moving syntax highlighting to the server


zeke:

zeke:
This is done!!!

PR #312, opened 2014-12-03
do syntax highlighting on the server

zeke:
Work in progress, following up on https://github.com/npm/newww/issues/78

This works for readmes on package pages, but the client-side highlighter is still doing work on other pages too, like `/enterprise`. Those are `hbs` files, so they don't go through a markdown processor before being rendered. 

One solution is to markdown-process all the templates before they are rendered.


zeke:
Closing, as this is superceded by https://github.com/npm/newww/pull/480

PR #325, opened 2014-12-04
R E G I S T R Y 2

zeke:
This is the first step in allowing users to sign up for private modules!


zeke:
github why you gotta out me?

PR #329, opened 2014-12-05
use `json` query param value to traverse context objects

zeke:
In development (and in product for npm employees), it's possible to examine the view context of any page by adding a `json` param to the query with no value:

[/~shtylman?json](https://preview.npmjs.com/~shtylman?json)

This patch allows the value of `json` to represent a deep key in the object:

[/~shtylman?json=profile.meta](https://preview.npmjs.com/~shtylman?json=profile.meta)

### DIFF
--- a/test/services/registry.js
+++ b/test/services/registry.js
@@ -75,7 +75,7 @@ describe('browsing', function () {
     });
   });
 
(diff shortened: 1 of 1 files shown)


zeke:
This is ready for merge.

PR #333, opened 2014-12-05
Fixes :hover on links in the right sidebar.

bslayton:
This change makes it so that links inside of .box in the right sidebar on package's page turn "npmred" on `.box a:hover` and not on `.box:hover` which causes all links in the parent .box to be turned "npmred" which, imo is bad UX.


zeke:
All right I'm just gonna ship this change and we can iterate on the link styles.

bslayton:
:metal:

PR #372, opened 2014-12-10
Fix username and logout links position in header

akenn:
The username and info links in the header look a little weird. 

They currently look like this:
- screen shot 2014-12-09 at 8 01 27 pm
- screen shot 2014-12-09 at 8 02 14 pm

This PR changes them to look like this:
- screen shot 2014-12-09 at 8 03 37 pm
- screen shot 2014-12-09 at 8 04 15 pm

Also side note: I noticed the logout link doesn't appear on mobile when a user is logged in because of [this line of code](https://github.com/npm/newww/blob/master/assets/styles/_header.styl#L295). Is this the intended behavior?


akenn:
Closing because it looks like this was fixed by commit 66ec00

PR #382, opened 2014-12-10
Adds style for tables in readme section

mikaelbr:
The tables in the READMEs looked a bit off without any styling. I tried to add a style similar to the code block styles, with a #f5f5f5 background and some padding.

I don't know what kind of style you are going for here. Do you think the background looks OK, or should it simply have borders or no background at all?

---

### Before
tablebefore
---
### After
tableafter

### DIFF
--- a/assets/styles/readme.styl
+++ b/assets/styles/readme.styl
@@ -31,6 +31,13 @@
   p, li
     font-size 16px
 
+  th
+    font-weight bold
+
+  td, th
+    padding 0 5px
+    background #F5F5F5
+
   p
     margin 15px 0
 


numtel:
I came here to write a very similar PR. The screenshot looks very good. Anything is better than the current no-style.

fvdm:
I agree, the tables **need** styling asap!
They are barely readable.

zeke:
Shipping this and keeping the conversation going at https://github.com/npm/newww/issues/545

fvdm:
Thanks, @zeke !

PR #387, opened 2014-12-10
Adds support for anchor links and adds heading anchor links

mikaelbr:
_Congratulations on the new site, by the way!_

Anchor links (#some-slug) didn't seem to pass the sanitation step, so I added a conditional to support it.

This PR also adds "Github style" heading anchor links. This way people can refer to specific part of the README (a anchor-icon is visible on heading hover). See image below.

I don't know if you are interested in this feature, but it's something I've been missing in the readmes, so I thought I'd give it a try.

anchorlinks

### DIFF
--- a/assets/styles/readme.styl
+++ b/assets/styles/readme.styl
@@ -47,6 +47,20 @@
   pre
     overflow-x auto
 
+  h2,h3,h4,h5
+    position relative
+    &:hover .package-header-link
+      visibility visible
+
+  .package-header-link
+    position absolute
+    padding 0 10px
+    left -30px
+    text-decoration none
+    visibility hidden
+    i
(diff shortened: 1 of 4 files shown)


mikaelbr:
It seems this has been implemented a different way, so this can be closed, @zeke ?

zeke:
It has indeed. Thanks for noticing.

The readme changes we just deployed are proving to be a bit volatile for some packages, so we're rolling back for now. But yes, the deep links are coming! :)

mikaelbr:
Sounds good. It looked very nice the packages I saw. Good job.

zeke:
Thanks for the kind words; they're good to hear in these sad rollbacky moments.

PR #402, opened 2014-12-10
Folder name changing

rockbot:
the facets/services experiment, though noble, has ended.

The facets are just handlers, and the services are just models, so let's call them that. :-)

@zeke - any thoughts?


zeke:
I think this may have been idle for too long. Shall we scrap it, @rockbot?

PR #429, opened 2014-12-12
Make README header sizes closer in line with bootstrap

vkarpov15:
Heya,

I dig the new site, but finding it somewhat frustrating that h3-h5 are the same size in readmes, which makes things more difficult to read: see https://www.npmjs.com/package/kareem . This commit makes the header sizes match Bootstrap's: http://getbootstrap.com/css/#type


zeke:
Thanks for the PR @vkarpov15. I agree that the headings need work, but a real solution to this will require more effort. README authors are very idiosyncratic in their choice heading hierarchies. Rest assured I will be working on this more in the near future.

PR #434, opened 2014-12-14
Adds server side initial npm expansion

sudodoki:
To prevent text change after initial page load by itself, suggest doing it server side.


zeke:
I'm gonna stick to my guns on this and say that `node package manager` is probably a good baseline, mostly for crawlers. Thanks for the PR though, @sudodoki.

PR #441, opened 2014-12-16
Ensuring license URL is valid, adding CC0 url. Fixes #440

### LINKED ISSUE #440: Broken link with 'CC0-1.0' license
When a package such as https://www.npmjs.com/package/express-secure-only sets it's license to `CC0-1.0`, the NPM page ends up with a broken link to http://opensource.org/licenses/CC0-1.0

I think the link *should* point to http://creativecommons.org/publicdomain/zero/1.0/ but no link at all might be good enough for things that aren't listed on opensource.org.

nfriedly:
Sets the license url to null if the license isn't listed on opensource.org or in the alternative licenses list I added (currently just CC0).

Updated the template to just display the name of the license without a link if no URL is available.

On my laptop, 11 tests failed both before and after this commit, so no new failures. The VM is still downloading but I'm 80% sure that it will work ;)

### DIFF
--- a/facets/registry/presenters/package.js
+++ b/facets/registry/presenters/package.js
@@ -8,7 +8,8 @@ var marked = require('marked'),
     similarity = require('similarity'),
     gh = require('github-url-to-object'),
     cheerio = require('cheerio'),
-    log = require('bole')('registry-package-presenter');
+    log = require('bole')('registry-package-presenter'),
+    osiLicenses = require('osi-licenses');
 
 module.exports = function package (data, cb) {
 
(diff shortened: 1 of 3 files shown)


zeke:
Thanks for the patch @nfriedly.. unfortunately much has happened in the codebase since you submitted it. Sorry about that. I recently extracted the OSS license URL generation into a standalone package: https://github.com/npm/oss-license-name-to-url -- this is being used on the r2 branch where major development is currently under way. If you're willing to port your fix over to that package, it would be much appreciated!

nfriedly:
Alright, sure

nfriedly:
OK, I copied everything except the `templates/registry/package-page.hbs` change to the new module. I'll submit a separate PR to the r2 branch here in a minute.

nfriedly:
Closing, this is now covered in other PRs.

PR #461, opened 2014-12-23
re-add opensearch.xml

dannycoates:
Before:

screen shot 2014-12-14 at 4 20 17 pm

After:

screen shot 2014-12-22 at 6 49 21 pm
screen shot 2014-12-22 at 6 50 15 pm

I stuck it in `assets/misc` since it seemed like the right place for it


zeke:
LGTM

zeke:
Thanks, @dannycoates!

:cake:

PR #473, opened 2014-12-29
Preserve query parameters in links when rendering README.

eiriksm:
Hey guys!

Great work on the new website!

Just noticed on a package I was browsing (karma at the time) that URLs with query parameters gets their parameters stripped. Not sure if you do it on purpose, but reading the code, it seems the query parameter is supposed to be preserved. So here is a PR fixing that, including a test (which will fail without the fix applied, to prove my point).

Basically, this is the problem:

```
$ node
> var u = { protocol: 'http:',
...   host: 'www.youtube.com',
...   pathname: '/watch',
...   search: 'v=dQw4w9WgXcQ',
...   hash: null }
undefined
> u
{ protocol: 'http:',
  host: 'www.youtube.com',
  pathname: '/watch',
  search: 'v=dQw4w9WgXcQ',
  hash: null }
> url.format(u)
'http://www.youtube.com/watch?v=dQw4w9WgXcQ'
> u.query = u.search
'v=dQw4w9WgXcQ'
> delete u.search
true
> url.format(u)
'http://www.youtube.com/watch'
>
```

Yup. So keep up the good work && <3 npm

Oh and a nice holiday and new years to npm and contributors!

### DIFF
--- a/facets/registry/presenters/package.js
+++ b/facets/registry/presenters/package.js
@@ -123,7 +123,7 @@ function urlPolicy (pkgData) {
         protocol: v.protocol,
         host: v.host,
         pathname: v.pathname,
-        query: u.query_,
+        search: u.query_,
         hash: u.fragment_
       }
     } else {
@@ -131,7 +131,7 @@ function urlPolicy (pkgData) {
         protocol: u.scheme_ + ':',
         host: u.domain_ + (u.port_ ? ':' + u.port_ : ''),
         pathname: u.path_,
-        query: u.query_,
+        search: u.query_,
         hash: u.fragment_
       }
     }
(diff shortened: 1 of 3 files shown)


zeke:
Thanks for the fix! However, we're about to switch to a new thing for readme processing (see https://github.com/npm/newww/pull/480), so I'm going to close this. To make sure your specific case is covered and won't regress, I ported your test over: https://github.com/zeke/marky-markdown/commit/0a1d3a51d4905d2ca7f87d26a8d05d9b66ed460f

:cake:

eiriksm:
Sounds good. And a :bow: for using the same video link
:chart_with_upwards_trend:

PR #480, opened 2015-01-02
marky-markdown

zeke:
This updates newww to use [marky-markdown](https://www.npmjs.com/package/marky-markdown) to process readme content. Highlights include:

- Human-readable code!
- Lots of tests
- Explicit HTML content policy with sanitize-html
- Server-side syntax highlighting
- Gravatar URL cleanup
- GitHub relative link cleanup
- Better tagging of badge elements
- Forward-looking CDN image support

This patch removes client-side syntax highlighting, so "static" pages like [npmjs.com/policies/privacy](https://www.npmjs.com/policies/privacy) will also need be passed through marky-markdown. That should be done before this is merged. :)


zeke:
This branch is up on staging now. Please help smoke test it by hitting a bunch of package pages and reporting any crashes, readme irregularities, or blemishes...

https://staging.npmjs.com/package/minimatch
https://staging.npmjs.com/package/cheerio
https://staging.npmjs.com/package/johnny-five
etc

zeke:
Latest thingies:

- deep link `#` symbols are hidden by default, displayed on hover
- paragraphs, lists, code blocks now all have no padding on the top and a little on the bottom. This way they all look well-spaced regardless of what precedes and follows them.
- code blocks have padding even if they don't get auto-identified and tagged as `hljs`

zeke:
@rockbot they are the same size in the css, so it seems to depend on the font. On my machine, they are the same size. The code font is `consolas`:

screen shot 2015-01-12 at 5 42 35 pm

Can you install http://chengyinliu.com/whatfont.html and let me know what font is being used on your machine?

zeke:
Ready!

Issue #479, opened 2014-12-30
Dockerize npm-www, running it using Docker / Fig


marcellodesales:
Hi there,

I'm been evaluating the integration of npm-www with the Enterprise server as a hosted solution. As we are dockerizing our internal assets, I have been trying to dockerize the setup of the npm-www server using Docker and Fig using the "dev/go.js" method.

The current demo of the server running is at https://asciinema.org/a/15074. I'm planning to submit a pull request if there are interests on this... 

Questions
-----
* Is the "dev/go.js" the ultimate way of deploying the server? 
* Are all those steps required when running a production setup of the Enterprise NPM?
* Is the CouchDB server in "dev/go.js" setup the same as the one running on the Enterprise NPM?

I can certainly remove/adjust the steps so that running the npm-www could be easily integrated with a Dockerized version of the Enterprise NPM...

thanks!
Marcello

marcellodesales:
Could someone explain the reason why replicate would fail?

```
npmRegistryRedisServer_1         | [1] 02 Jan 22:44:24.549 - 0 clients connected (0 slaves), 437000 bytes in use
npmregistryserver_1              | Running "/usr/local/bin/node" ["/usr/src/app/dev/replicate.js"]
npmRegistryRedisServer_1         | [1] 02 Jan 22:44:29.559 - 0 clients connected (0 slaves), 437000 bytes in use
npmregistryserver_1              | Running "/usr/local/bin/node" ["./node_modules/npm2es/bin/npm2es.js","--couch=http://192.168.4.201:15984/registry","--es=http://192.168.4.197:9200/npm"]
npmregistryserver_1              | did npm2es undefined
npmregistryserver_1              | 
npmregistryserver_1              | TypeError: Cannot read property '_rev' of undefined
npmregistryserver_1              |     at Request._callback (/usr/src/app/dev/replicate.js:81:21)
npmregistryserver_1              |     at self.callback (/usr/src/app/node_modules/request/request.js:373:22)
npmregistryserver_1              |     at Request.emit (events.js:95:17)
npmregistryserver_1              |     at Request.onRequestError (/usr/src/app/node_modules/request/request.js:971:8)
npmregistryserver_1              |     at ClientRequest.emit (events.js:95:17)
npmregistryserver_1              |     at Socket.socketErrorListener (http.js:1551:9)
npmregistryserver_1              |     at Socket.emit (events.js:95:17)
npmregistryserver_1              |     at net.js:833:16
[...]

PR #485, opened 2015-01-03
Dev and Deploy the server using Docker and Fig

### LINKED ISSUE #965: "sign up for private modules" still shows up on profile page for users with private modules
1) Sign in to an account that has already paid for npm private modules.
2) Go to profile page.
3) See, directly under photo, "sign up for private modules".

It should probably say something else for those who have already signed up ;-)

marcellodesales:
What
----
This pull request is to solve the feature request that allows developers to use Docker to locally develop and deploy a complete instance of the server, with replication.

Fixes #965 as described below.

How
-----
Instead of requiring Vagrant, all depending services and databases are created using Docker Data Containers and servers, along with the build of the current server's code as a Docker container.

Description
----
The architecture of this build service is as follows:

* Docker: Container technology.
 * Uses a `Dockerfile` to describe the steps to build a new image. A parent/base image is used from the "trusted" docker hub. 
* Fig: How to describe multiple containers and their relationships.
 * Uses a `fig.yml` file to describe multiple containers and how they relate to each other.

The server was created using the following design-patters:

* Data Container: A container that knows `where` to store data in the file-system.
* Database Server Container: A container that spawns the process of a Database server, and uses a Data Container for persistence. Those can use the "trusted" official images deployed by each vendor on Docker hub.
* Server Container: The container that builds and serves the HTTP server in any technology. Images are usually created using the "trusted" official image for the technology used, in our case Node.js.

Fig.yml defines the 3 data containers for all the depending data stores: couch, elastic search and redis. Similarly, it also defines a single server based on Node.js image as mentioned.


zeke:
See https://github.com/npm/newww/issues/479#issuecomment-75646269

PR #500, opened 2015-01-06
updates explicit installs data

rockbot:
unfortunately this needs to be done by hand atm :speak_no_evil: 

cc: @zeke


zeke:
I see all the same "packages people 'npm install' a lot" on the home page. Is that expected?

PR #503, opened 2015-01-06
the project does not use a help wanted label, it uses patch welcome

faiq:
this changes "Help Wanted" to "patch welcome"
cc: @zeke @rockbot


faiq:
lol someone got this in after i fixed it already. closing

PR #504, opened 2015-01-07
fix event tracking for hiring ads

zeke:
@rod11 discovered that custom event tracking stopped working when we launched newww. This PR is an attempt to fix that.

@rockbot can you take a look?


zeke:
:postal_horn:

PR #507, opened 2015-01-07
Upgrade to hapi 8.1

rockbot:
adds some more to https://github.com/npm/newww/pull/502 - plus this way we can push it up to staging ;-)


zeke:
Wow! Thanks for the help, @hueniverse.

Issue #476, opened 2014-12-30
Upgrade vagrant box and/or docs?


eiriksm:
Hello.

While preparing for submitting a patch today, I encountered a series of hurdles which I had to overcome to get the local development up and running. This issue will serve as a summary of that, and I'd be happy to create separate issues (and PRs) for things that needs to be fixed. This is for discussion.

Let me just say that my motivation for writing this is that I think it is awesome that we can all contribute to the npm website, and it is awesome that a vagrant box exists to get up and running from everywhere. And of course npm is on top of the Niceness Per Minute(tm) scale. My motivation is to make the first boot experience great too.

First:
when following the exact steps in the readme, the web site does not start. In fact, it does not install. This is probably the same issue as @nfriedly is having over at #441.

A little snip from the log is this:
```
> hiredis@0.1.17 install /vagrant/node_modules/hiredis
> node-gyp rebuild

make: Entering directory `/vagrant/node_modules/hiredis/build'
  CC(target) Release/obj.target/hiredis/deps/hiredis/hiredis.o
  CC(target) Release/obj.target/hiredis/deps/hiredis/net.o
  CC(target) Release/obj.target/hiredis/deps/hiredis/sds.o
  CC(target) Release/obj.target/hiredis/deps/hiredis/async.o
  AR(target) Release/obj.target/deps/hiredis.a
  COPY Release/hiredis.a
  CXX(target) Release/obj.target/hiredis/src/hiredis.o
make: g++: Command not found
```

OK, so no g++. This is solved by simply running `sudo apt-get install g++`. 

Also, node version is 0.10.26. Which is not the most recent one :p

There were other small problems as well, but this brings me to my points:
- I have been working alot with vagrant, and would love to add these fixes to the vagrant box. But where would I post PRs to the box? And how often does it get built and posted?
- I could just add some extra lines to the Vagrantfile to make sure node is updated, along with installing g++. But then we are updating an outdated version of node instead of just installing the newest one. Which is preferred?
- I could add some lines in the docs about "install g++" and so on, but that feels almost like excuses for not doing it right in the first place. If you see what I mean.

Ok. That's a start. Let's make the "get up and running" with the npm website feel like "No Problems Met"(tm)

Have a great day!

zeke:
Thanks for taking the time to write up this issues. Here's the TLDR:

- A compiler is required
- Node needs to be updated
- Readme should reflect this stuff

I'll defer to @seldo and @rockbot for more help on this. They're the VM experts. :)

[...]

PR #524, opened 2015-01-10
Make vagrant box start from base ubuntu.

eiriksm:
This is the PR generated from discussion over at #476

Discussion points I can think of:
- Ubuntu has been upgraded to 14.04. Good/bad?
- Couchdb has been upgraded to 1.6. As suggested. Good?
- Elasticsearch is not upgraded (still 1.1). Good/bad?
- Redis is a slightly different version, but follows my go-to ppa for it. Good/bad?

Instructions from README still applies with this setup, should not change anything as far I as I know.

So, up for discussion is the keyword here :)

And have a nice weekend!

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -10,11 +10,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # please see the online documentation at vagrantup.com.
 
   # Every Vagrant virtual environment requires a box to build off of.
-  config.vm.box = "npm-www-20140617"
-
-  # The url from where the 'config.vm.box' box will be fetched if it
-  # doesn't already exist on the user's system.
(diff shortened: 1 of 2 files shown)


gtanner:
Hey ...

didn't see this before I issued my PRs (https://github.com/npm/newww/pull/615 and https://github.com/npm/newww/pull/614).

There is a couple of extra things I have in mine that are missing from this:

- creation of the default config files (if they don't exist)
- auto running of ```npm install```
- handle the missing redis password in config.

Not sure which PR's should be the ones to go with. I do like this PR a little better as there is less magic in the vagrant box that get's pulled down.

zeke:
@eiriksm thanks for the PR, but we probably won't need it in the very near future: https://github.com/npm/newww/issues/761

Issue #508, opened 2015-01-07
Show GitHub contributor count on the package page


zeke:
I find GitHub contributor counts to be insightful when assessing project "health". For packages with GitHub-based repositories, we could easily add a contributor count to the stats, just like we do with open issue and PR counts.

GitHub API doc: https://developer.github.com/v3/repos/#list-contributors

Where the code should go: https://github.com/npm/newww/blob/master/assets/scripts/github.js

This is a small, contained feature, and could be a good opportunity for someone in the community who wants to contribute.

yashprit:
@zeke can I take this or is this something not planned yet

zeke:
@yashprit I know @coleGillespie started on it a while back, but the effort kind of fizzled out, which is understandable. The Github API code we're running client-side is in definite need of a cleanup. It's kind of spaghetti-ish right now. It would be helpful to use promises or a lightweight async tool to handle all those ajax requests more cleanly.

@yashprit if you do submit a PR, start small and get the conversation going early. I wouldn't want you to put a bunch of work into something we can't ship.

yashprit:
@zeke that make sense.

zeke:
Work in progress: https://github.com/npm/newww/pull/824

PR #526, opened 2015-01-10
added counts of contributors to github stats on package page for #508

coleGillespie:
followed current paradigm. may want to re order the list items on the right side of the page.

### DIFF
--- a/assets/scripts/github.js
+++ b/assets/scripts/github.js
@@ -12,6 +12,9 @@ github.init = function(){
   github.pull_requests = {
     element: $("#pull_requests")
   }
+  github.contributors = {
+    element: $("#contributors")
+  }
 
   // If there's an #issues element on the page, look for the repo's
   // GitHub API base URL in a data attribute
@@ -22,6 +25,8 @@ github.init = function(){
     // Set up API URLs
     github.issues.api_url = github.issues.element.data().ghapi+"/issues?per_page=1"
     github.pull_requests.api_url = github.issues.element.data().ghapi+"/pulls?per_page=1"
+    github.contributors.api_url = github.issues.element.data().ghapi+"/contributors"
+
 
     // Start with pull requests
     github.getPullRequests()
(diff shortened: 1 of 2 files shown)


zeke:
Awesome!

I see one potential snag. Currently the GitHub API base URL is added to the HTML page as a data attribute only if `bugs.url` is defined in package.json. I think we should move the `if ghapi` conditional out of `bugs` and into `repository` so it has a greater chance of being included in the template, because most package.json files define `repository`, whereas `bugs` is less commonly included.

Does that make sense, @coleGillespie?

coleGillespie:
@zeke got it, let me have another pass at this and see what i can do

coleGillespie:
@zeke is this what you were looking for?

zeke:
I think I led you astray. I was thinking the ghapi data attribute should be attached to the repository element, as opposed to the contributors or issues elements. But now that I think about it this may complicate things. I would revert that data attribute change and leave it where it was.

Another thing: The GitHub API returns 30 results by default, and a max of 100. But popular projects have well over 100 contributors (express has 166). There's a little trick you can do with the GitHub API to get around this: ask for one single result in your API call, then look at the response headers to find out how many "pages" there are. That's what we're currently doing to get issue/PR counts:

```js
Number(parseLinkHeader(xhr.getResponseHeader("Link")).last.page)
```

https://github.com/npm/newww/blob/1e0ebe7158f982e9fbd3ac0a46086d80e2415346/assets/scripts/github.js#L50

coleGillespie:
https://api.github.com/repos/joyent/node/contributors

zeke:
Try this:

```sh
curl -i https://api.github.com/repos/joyent/node/contributors?per_page=1
```

In the response headers, you'll see:

```
Link: <https://api.github.com/repositories/211666/contributors?per_page=1&page=2>; rel="next", <https://api.github.com/repositories/211666/contributors?per_page=1&page=548>; rel="last"
```

The `last` value (548) is the number of contributors

coleGillespie:
ha, ok lets see what i can do

zeke:
@coleGillespie I'm gonna put this on ice. Feel free to re-open if you end up working on it more. Thanks!

PR #533, opened 2015-01-12
remove unused dependency

zeke:
https://www.npmjs.com/package/accept-language-parser was being used to figure out the user's language and preferred big-number delimiter, but now we're using the browser's built-in `toLocaleString()` function instead. Yay!


zeke:
Closed this because on second thought, we kinda need this package anyway to do localization. :)

PR #539, opened 2015-01-13
Added border-radius to the ad class

RichardLitt:
The only other elements that don't have border-radii on the packages page are the code blocks; I've added a border-radius as a further way of differentiating the ad, in addition to the current border.


zeke:
I'm blindly merging this because you used the word radii

RichardLitt:
:D

Issue #548, opened 2015-01-15
Add autofocus to the login form


bmullan91:
Speaks for itself really. I saw issue #316, but I think in this scenario adding the autofocus is a valid use case.

zeke:
Great idea. Patches welcome!

bmullan91:
Closing this - autofocus FTW!

PR #549, opened 2015-01-15
adding auto focus issue #548

bmullan91:


andineck:
+1

aredridel:
autofocus wrecks accessibility, but you can type "/" to focus the field on the page!

andineck:
thx, for the tip, didn't know. Is that a common thing working on different websites?

aredridel:
Many, particularly sites written by unix using geeks. Also, tab can select the first thing in many browsers.

andineck:
:-) cool, will check, only tried tab so far.

PR #558, opened 2015-01-19
add npm weekly to menu

### LINKED ISSUE #554: Add a link to the npm weekly on the homepage
@rod11 suggested we have something on the homepage to point people to the npm weekly. I'm thinking maybe a box of some sort?

In the meantime, we could add it to the menu.

linclark:
Fixes #554


zeke:
We definitely need to handle the layout weirdness in the murky space between phone and full-screen "desktop", but that shouldn't block this. Gonna ship it.

PR #563, opened 2015-01-20
add npm weekly subscription opt-in to sign up

### LINKED ISSUE #555: Add opt-in subscription to npm weekly during sign up
When people are signing up on the site, there should be an opt-in checkbox to add them to the npm Weekly.

linclark:
Will fix #555 when it's done.

I figured I'd take a shot :) This is a first pass at adding mailchimp to the sign up flow. The `xxx` in there is a placeholder for the auth token. Where do we put that kind of config?

Also, it doesn't do anything on success or error. I don't think we want to display anything to the user, but should we log errors?

### DIFF
--- a/services/user/methods/signupUser.js
+++ b/services/user/methods/signupUser.js
@@ -2,12 +2,20 @@ var Boom = require('boom'),
     Hapi = require('hapi'),
     anonCouch = require('../../../adapters/couchDB').anonCouch,
     metrics = require('../../../adapters/metrics')(),
-    mailchimp = require('mailchimp-api');
+    mailchimp = require('mailchimp-api'),
+    log = require('bole')('user-signup');
 
 module.exports = function signupUser (acct, next) {
   if (acct.npmweekly === "on") {
     var mc = new mailchimp.Mailchimp(process.env.MAILCHIMP_KEY);
-    mc.lists.subscribe({id: 'e17fe5d778', email:{email:acct.email}});
+    mc.lists.subscribe({id: 'e17fe5d778', email:{email:acct.email}}, function(data) {
+      // do nothing on success
+    }, function(error) {
+      log.error('Could not register user for npm Weekly: ' + acct.email);
+      if (error.error) {
+        log.error(error.error);
+      }
+    });
   }
 
   var timer = { start: Date.now() };
--- a/services/user/methods/signupUser.js
+++ b/services/user/methods/signupUser.js
@@ -6,7 +6,7 @@ var Boom = require('boom'),
 
 module.exports = function signupUser (acct, next) {
   if (acct.npmweekly === "on") {
-    var mc = new mailchimp.Mailchimp('xxx');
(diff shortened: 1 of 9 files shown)


services/user/methods/signupUser.js
```diff
@@ -1,9 +1,15 @@
 var Boom = require('boom'),
     Hapi = require('hapi'),
     anonCouch = require('../../../adapters/couchDB').anonCouch,
-    metrics = require('../../../adapters/metrics')();
+    metrics = require('../../../adapters/metrics')(),
+    mailchimp = require('mailchimp-api');
 
 module.exports = function signupUser (acct, next) {
+  if (acct.npmweekly === "on") {
+    var mc = new mailchimp.Mailchimp('xxx');
```
rockbot: for this, we can add an environment variable in the .env file (see .env.example)... we'll also need to ad `require('dotenv').load()` to the top of server.js :-)

zeke: If we're adding a secret here, it'll need to go into neww-config too, in the `env` stanza:

- https://github.com/npm/newww-config/blob/master/production/service.json
- https://github.com/npm/newww-config/blob/master/staging/service.json

zeke:
I wish we parsed locale strings as markdown. That way you could do "Sign me up for [npm Weekly](...) emails". But that's of course not a blocker.

services/user/methods/signupUser.js
zeke: This `xxx` should be `process.env.MAILCHIMP_API` or similar.

rockbot: Yes! What @zeke said :-D

zeke:
@linclark are you comfortable adding some tests for this? If not, I can help.

linclark:
Forgot to mention, this could possibly use some CSS adjustments.

npm

zeke:
Try putting a div around it with `.radio-group` class:

screen shot 2015-01-20 at 4 29 13 pm

And feel free to rename radio-group to something more general :)

zeke:
PS that screenshot above is from https://www.npmjs.com/contact, which incidentally has some oddly formatted humans right now. So many css regressions. So little time!

linclark:
npm

Thanks for the tip, looks good to me :)

linclark:
Ok, I ended up going with Sinon's spy. I modified a suggestion from CJ and basically added something analogous to setter injection. I also tried writing the test with nock and left the commit there in case folks want to look.

I wasn't able to add the test as part of `signup.js` since that mocks out the actual `signupUser` method, which is what I wanted to test. Instead, I added the tests to the user service tests. 

It currently only tests that the call to Mailchimp is made based on the checkbox, but it would be possible to test whether the users are created as well.

zeke:
Tests passing; works in the browser; looks good to me!

One suggestion: add `MAILCHIMP_KEY=12345` to `.env.example`

:shipit:

PR #565, opened 2015-01-20
Changes for translate

alejonext:
Well I made a number of changes to make the translation is more simple.


zeke:
@alejonext thanks for your herculean effort here. Unfortunately we won't be able to merge this. It is not entirely in line with the vision @rockbot and I have for localizing the site, and too much has changed in the codebase since you submitted the PR.

That said, we are open to a more bite-size approach. We will be merging the long-lived `r2` branch to `master` very soon, after which we will happily discuss a localization strategy with you and whoever else wants to help. I hope you understand, and again thanks for your contribution. It is appreciated.

alejonext:
@zeke OK !!! I would be happy to discuss this strategy!

zeke:
Thanks for your understanding, and stay tuned :)

PR #568, opened 2015-01-21
add Schema.org structured data for organization

linclark:
Using Schema.org can provide [better looking listings](https://developers.google.com/webmasters/structured-data/customize/social-profiles) in the search engines, especially Google.

### Testing instructions
Structured data can be tested with https://developers.google.com/webmasters/structured-data/testing-tool/

### DIFF
--- a/templates/company/index.hbs
+++ b/templates/company/index.hbs
@@ -87,3 +87,14 @@
     <div class="hiring-container" data-template="vanilla"></div>
   </li>
 </ul>
+<script type="application/ld+json">
+  { "@context" : "http://schema.org",
+    "@type" : "Organization",
+    "name" : "npm, Inc.",
+    "url" : "https://www.npmjs.com",
+    "sameAs" : [ "http://twitter.com/npmjs",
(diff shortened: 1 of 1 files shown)


zeke:
Does the script tag generate a content security warning in the browser?

Why only on this page? Is that the convention?

linclark:
I don't think that it generates a content security warning. I think that only happens when there's transclusion going on, so for things like `src` on a script or `href` on a link, where the client is making a request across the network to include another resource. This JSON will be read by the search engine's parser, but won't do anything on the client.

I'm pretty sure it only needs to be on one page (Google says it "can be inserted into any HTML page on your company's official website that is crawled and indexed by Google") so I went with the one I thought would be crawled most often.

zeke:
> In computer science, **transclusion** is the inclusion of a document or part of a document into another document by reference. For example, an article about a country might include a chart or a paragraph describing that country's agricultural exports from a different article about agriculture.

I guess I dropped out before we got to that word. :)

PR #572, opened 2015-01-22
access page

zeke:
WORK IN PROGRESS. DO NOT MERGE.

access-1

access-2

access-3


seldo:
I like the third option best visually but think changing the labels in it back to "read-only" and "read-write" would be less ambiguous.

"Remove" is definitely clearer than "revoke".

zeke:
Microprogress...

screen shot 2015-01-26 at 5 05 13 pm

PR #573, opened 2015-01-22
Marky Markdown, Async Edition

zeke:
Last week we deployed marky-markdown to www production and it broke stuff and caused 503s because the highlight.js syntax highlighter was hanging when parsing packages with weird content. Some of those packages:

- http://npm.im/flake
- http://npm.im/grunt-angular-templates
- http://npm.im/cicada
- http://npm.im/mkhere

So I revised marky-markdown to use pygmentize-bundled instead of highlight.js. Pygments is the syntax highlighter used by GitHub, and seems to be more widely used and maintained than any other syntax highlighter out there.

pygmentize-bundled shells out to the pygments python binary and has an async API, so marky-markdown's function signature had to be updated to be async. Details here: https://www.npmjs.com/package/marky-markdown

Syntax highlighting using pygments is working much better, and it's not choking on all the packages that were making highlight.js choke, BUT it's still pretty slow. When parsing long READMEs like https://www.npmjs.com/package/async#readme, it takes several seconds.

This pull request includes the new asynchronous version 2 of marky-markdown, with the `highlightSyntax` option set to `false`. This means code snippets will not have syntax highlighting for the time being, but we'll get all this good stuff:

- Removes broken and malicious input like script tags and iframes using [sanitize-html](https://www.npmjs.com/package/sanitize-html)
- Turns headings (h1, h2, etc) into deep links
- Rewrites relative GitHub links to their absolute equivalents
- Rewrites relative GitHub images sources to their absolute equivalents
- Rewrites insecure Gravatar URLs to use HTTPS
- Wraps embedded YouTube videos so they can be styled.
- Removes redundant content that closely matches npm package name and description
- Parses `package.description` as markdown
- Applies CSS classes to badge images, so we can do something interesting with them one day.

This is running on staging now.

Package pages:

- https://staging.npmjs.com/package/cheerio
- https://staging.npmjs.com/package/async
- https://staging.npmjs.com/package/flake
- https://staging.npmjs.com/package/cicada

Static pages too:

- https://staging.npmjs.com/policies/privacy
- https://staging.npmjs.com/policies/receiving-reports


dev/replicate.js
```diff
@@ -49,7 +49,7 @@ var packagesByRockbot = ["couch-login", "googalytics", "newww",
 
 var crazyPackages = [
   'ifyouwanttogetthesumoftwonumberswherethosetwonumbersarechosenbyfindingthelargestoftwooutofthreenumbersandsquaringthemwhichismultiplyingthembyitselfthenyoushouldinputthreenumbersintothisfunctionanditwilldothatforyou',
-  'dezalgo', 'JSONStream'
+  'dezalgo', 'wzrd' 'dezalgo', 'JSONStream'
```
rockbot: ... dezalgo is here twice? :-)

PR #583, opened 2015-01-27
adds a .jshintrc to the project

rockbot:


zeke:
It looks like https://github.com/npm/newww/pull/272 has some extra goodies like an ignore file and a gulp task. Should we merge that first, then this?

Issue #513, opened 2015-01-08
focus search input when / key is pressed


zeke:
That's a convention, right?

We're already using mousetrap.js, so it should be trivial to add this feature.

coleGillespie:
@zeke can you explain further how this should work?

PixnBits:
I would suggest that on the index page the search box should be focused w/o needing a shortcut key pressed.
Other pages I like the shortcut.

@coleGillespie I'd guess something akin to
```javascript
Mousetrap.bind('/', function(e) {
    $('#site-search').focus();
});
```

zeke:
@PixnBits is on the right track with that snippet, though I disagree about the autofocus-on-homepage suggestion. We had it that way originally and people freaked out. Apparently people use arrow keys to scroll..

PixnBits:
Scroll? *Pfft* **Search** is where it's at :smile: 
Makes sense, though some indicator/reminder of the shortcut would be nice (at least for *my* use case :-S ).

I'm trying
```javascript
mousetrap.bind('/', function(e){
  $('#site-search').focus();
  // don't add '/' character to input element
  e.preventDefault();
});
```
and looks like its working. Not sure what the best file would be to add this...perhaps [index.js](https://github.com/npm/newww/blob/master/assets/scripts/index.js)?

Would we want to also place the cursor at the end of the existing text in the input (instead of the beginning)? Example, I already searched for 'oracle' and when I press <kbd>/</kbd> the cursor is put before the 'o'.

PixnBits:
ah, I see what you're saying with #316

silverwind:
I'd vote for just autofocus when navigating to the front page. Subsequent page loads should probably not focus. Maybe it's possible to let the arrow key events bubble up from the input to still retain scrolling?

PixnBits:
Sweet, the <kbd>/</kbd> shortcut [is live](https://www.npmjs.com/)! Feedback on your usage?

silverwind:
Works as expected, but as a user of a german keyboard, `/` is very hard to reach at`Shift` + `7` (Whoever designed that layout should burn in hell).

A convention I saw for this in other software is to also bind `-` which is at the same location the `/` is on an american keyboard. Would you consider taking a PR for this?

zeke:
> Would you consider taking a PR for this?

Sure... let's try it and if it causes problems for people we can revisit.

silverwind:
I'll research first if it's possible to rely on a key code or something for the key besides shift.

[...]

PR #602, opened 2015-01-30
focus search input when / key is pressed (#513)

PixnBits:
as per discussion [thus far](https://github.com/npm/newww/issues/513#issuecomment-70174406)


zeke:
:+1: thanks

PR #609, opened 2015-02-02
R2 environment

zeke:
Here's what you see when you're missing stuff:

screen shot 2015-02-02 at 11 50 50 am


zeke:
@rockbot because this change breaks tests! If you want, we can turn this into the "fix test env" branch and iterate..

zeke:
screen shot 2015-02-02 at 3 36 07 pm

PR #614, opened 2015-02-03
Add missing default password for redis

gtanner:
Since we configure redis by default to have a password, this config file
should also have that password. (see: https://github.com/npm/newww/blob/master/dev/redis/redis.conf#L16)

I was getting hit by this when attempting to follow the README.md setup instructions.


zeke:
Yeah let's take it out of `redis.conf` instead. Mind updating the PR?

zeke:
This is solved in `r2` branch, which will land any day now. Closing.

PR #615, opened 2015-02-04
Improve vagrant up experience.

### LINKED ISSUE #611: Vagrant vm is missing g++
In attempting to start up a new local dev environment the call to:

```
npm install
```

fails with a native build error

Running:
```
sudo apt-get install g++
```

in the VM and rerunning ```npm install``` will now work.

gtanner:
This change will improve the setup of new developers and make the
initial clone and configure much easier.

Now to get the system in a state to be able to run will just require a
single call to `vagrant up`

- copy config files from examples (if they do not exist already)
- fixes #611 (temporary) to install g++ on the vm
- runs npm install (as the vagrant user)


zeke:
@gtanner thanks for this PR, but we'll soon be moving away from the vagrant image. Details at https://github.com/npm/newww/issues/761

PR #616, opened 2015-02-05
add a google verification route

zeke:
screen shot 2015-02-04 at 5 39 57 pm


zeke:
@seldo this is live: https://www.npmjs.com/google17836d108133913c.html

PR #619, opened 2015-02-05
Update style to solve issues on last publisher 

### LINKED ISSUE #550: "User last published" sidebar line height borked
On mobile chrome, that is.

elmasse:
Fixes #550. The last publisher info is shown underlined on Safari browsers (both mobile and OS X)


zeke:
Thanks @elmasse! Landed in https://github.com/npm/newww/pull/831

PR #633, opened 2015-02-09
bump hiredis to 0.2 for iojs compatiblity

zeke:
cc @seldo


seldo:
This 3-character edit is now officially approved :+1:

PR #636, opened 2015-02-11
collaborators

zeke:
This PR adds the `Collaborator` model, routes, handlers, fixtures, mocks, and tests. The UI is not yet complete, but this is ready for merge into r2, pending any problems uncovered by @rockbot or @bcoe 

@rockbot I went snooping in the [crumb source](https://github.com/hapijs/crumb/blob/master/test/index.js#L103) and discovered the origin of the token generation technique we're using in tests now. I created [test/handlers/crumb.js](test/handlers/crumb.js) as a simpler abstracted way to get a csrf token in tests that doesn't add a bunch of extra code to each test. Eventually we should use this helper in all the tests that are doing it [the long way](https://github.com/npm/newww/search?utf8=%E2%9C%93&q=x00&type=Code).

@bcoe now that I have all this integration work done, I can finally start implementing the UI. However, I've had a realization: I need more user metadata for each collaborator: things like email (so we can generate the gravatar URL), full name, etc. Any ideas on how to get this extra data without making a request to user-acl for each collaborator? Maybe a batch user endpoint?


zeke:
Another option would be to make the collaborators endpoint return more user properties... not sure if that's any easier/better.

bcoe:
@zeke I've updated the collaborators API so that you now get detailed user information.

PR #640, opened 2015-02-12
(also) Bind search shortcut to '-' (easier on non en-us keyboards)

PixnBits:
per @silverwind's https://github.com/npm/newww/issues/513#issuecomment-72329897

### DIFF
--- a/assets/scripts/index.js
+++ b/assets/scripts/index.js
@@ -26,7 +26,7 @@ $(function () {
   })
 
   // Focus search input when / key is pressed (#513)
-  mousetrap.bind('/', function(e){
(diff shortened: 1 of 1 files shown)


silverwind:
Thanks!

I haven't gotten around to exploring whether it's possible to find the correct key besides the right shift. Shifting through the [various layouts](http://en.wikipedia.org/wiki/Keyboard_layout), `.` also seems to be pretty common on that position, but I guess you'd have to draw a line at some point :)

zeke:
This is up on www.npmjs.com now. Give it a try!

silverwind:
Works, thanks!

PR #646, opened 2015-02-17
A more restrictive CSP for Hubspot

zeke:
From @jlamendo:

> the hubspot CSP urls are not restrictive enough, and could cause a bypass if a less restrictive injection point were to be found. I know it results in adding to an already lengthy CSP header, but you should whitelist specific pages on hubspot, or at least only allow pages located within your client ID folder.

It would be nice to factor out the hubspot config from `config.js` into its own module like `lib/hubspot.js` and stop passing config through the handlers, but it's tangled and I don't want to go down a rabbit hole this morning.

### DIFF
--- a/lib/csp.js
+++ b/lib/csp.js
@@ -4,18 +4,24 @@ csp.default = {
   defaultSrc: 'self',
   scriptSrc: [
     'self',
-    'https://checkout.stripe.com',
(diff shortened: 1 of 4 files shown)


zeke:
@ceejbot I locked it down a bit more by changing `https://checkout.stripe.com/` to `https://checkout.stripe.com/checkout.js` but I think that's the best we can do.

Issue #656, opened 2015-02-19
Package page new nav


ceejbot:
New navigation for new features. Description somewhat vague.

zeke:
Closing this in favor of the PR

PR #659, opened 2015-02-19
[WIP] r2 package nav

zeke:
Picking up where https://github.com/npm/newww/issues/656 left off.

The idea is this: When viewing a package at `/package/foo`, a nav should be displayed to all users that have collaborative access to that package. Currently the only things linked to by the this nav will be:

- `/package/foo`
- `/package/foo/access`

Eventually there will be other links in the nav, so that should be taken into consideration.

For users who are not collaborators on the package (almost everyone), the nav element will simply not exist.


zeke:
This update to marky-markdown needs to be landed before I can proceed: https://github.com/npm/marky-markdown/pull/29 -- this is pretty much a breaking change, which would put marky-markdown at [five point oh](https://image.adam.automotive.com/f/54267105/1306-1989-ford-mustang-5-0-lx-side-passenger-view).

PR #660, opened 2015-02-19
nix recent authors

zeke:
https://www.npmjs.com/recent-authors is:

- broken. Clicking on author names leads to `/package/authorname`... WAT
- not hyperlinked from anywhere on the site
- unpopular, with only [9 pageviews](https://cloudup.com/cjUcv2d8NcT) in the last thirty days.

This PR removes the `/recent-authors` path and redirects requests to `/`. It also adds some tests around various redirects we're already doing.


zeke:
For the record, this is not about me being opposed to having this kind of feature on the site in the future. We just don't currently have this endpoint slated for registry 2, and no one has complained that it's been gone for nearly three months.

isaacs:
Well, it's clearly broken, and like you said, no one's using it.  LGTM, but suggest also adding an issue to show recently active users on the site somewhere, so we can get it on our roadmap, even as a modest-priority thing.  No one views this page any more, but that's mostly because it used to be on the homepage and now it isn't.  I remember people used to tweet about seeing their name on the homepage, and it seems a bit sad that this isn't a thing we have any more.

PR #680, opened 2015-02-26
add schema.org metadata to package pages

### LINKED ISSUE #625: Structured metadata for better search results
Looking at our site in Google's Structured Data Testing Tool it appears Google has trouble interpreting our site as a content site and is instead interpreting it as a search engine:

http://www.google.com/webmasters/tools/richsnippets

We can probably rectify this by adding metadata that declares package pages as "products" and giving them explicit descriptions, etc.

linclark:
Fixes #625 

The applicationCategory and offers values are a little weird. I added them because you're required to have 2 of 4 optional fields, and those two were the closest to making sense for our use case (the other two are aggregateRating and operatingSystem).

This could potentially go in `package-page.hbs`, but I wasn't sure whether canonicalURL was/should be exposed as a variable there.

### DIFF
--- a/templates/registry/package-page.hbs
+++ b/templates/registry/package-page.hbs
@@ -166,6 +166,22 @@
     </p>
   {{/if}}
 
+  <script type="application/ld+json">
+    {
+      "@context": "http://schema.org",
+      "@type": "SoftwareApplication",
+      "name": "{{name}}",
+      "description": "{{description}}",
+      "url": "{{../canonicalURL}}",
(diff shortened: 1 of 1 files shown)


zeke:
`canonicalURL` is set on all pages. Might we get dinged for including this on all pages? If not, I'm cool with shipping this as-is.

linclark:
Discussed this with @zeke. `canonicalUrl` is not a problem, but it was possible to move this to `package-page.hbs`, which is a better place for it to live.

PR #683, opened 2015-02-26
Continuous date formatting

zeke:
Update timestamps on the page every few seconds


zeke:
`setInterval` is a famous way to introduce memory leaks into an app, but here we're not doing anything crazy or mysterious. So the answer is... not that I know of. :shrugs:

zeke:
Backed off to 10 seconds. This is ready for merge.

PR #684, opened 2015-02-26
R2 package downloads handling

zeke:
Moved the package page 404 `catch` block up above the downloads API call, so that can be caught independently.

### DIFF
--- a/handlers/package.js
+++ b/handlers/package.js
@@ -15,6 +15,24 @@ package.show = function(request, reply) {
   request.logger.info('get package: ' + name);
 
   var promise = Package.get(name)
+    .catch(function(err){
+      if (err.statusCode === 404) {
+        if (validatePackageName(name).validForNewPackages) {
+          context.package = {name: name}
+          request.logger.error('package not found: ' + name);
+          reply.view('errors/not-found', context).code(404);
+          return promise.cancel();
+        }
+
+        request.logger.error('invalid package name: ' + name);
+        reply.view('errors/not-found', context).code(400);
+        return promise.cancel();
+      }
+
+      request.logger.error(err);
+      reply.view('errors/internal', context).code(500);
+      return promise.cancel();
+    })
     .then(function(p) {
       package = p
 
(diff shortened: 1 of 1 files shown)


handlers/package.js
```diff
@@ -15,6 +15,24 @@ package.show = function(request, reply) {
...
+        }
+
+        request.logger.error('invalid package name: ' + name);
+        reply.view('errors/not-found', context).code(400);
+        return promise.cancel();
+      }
+
+      request.logger.error(err);
+      reply.view('errors/internal', context).code(500);
+      return promise.cancel();
```
rockbot: what does `promise.cancel()` do?

zeke: It breaks out of the promise chain. I found out about it after the tests were yelling at me for calling `reply.view` more than once.

zeke: Maybe @ceejbot or @bcoe know of a more elegant way to do this? I didn't like having to do `var promise = Package.get(...)`

bcoe: In some cases you want to ignore the error and trickle up to the next `then`, but in other cases you want to abort all the things?

zeke: Any time we reach a `reply.something`, I think it's appropriate to cancel the promise.

bcoe: Gotcha. Here's an idea for a refactor that would clean stuff up., but is not required for this pull:

* pull the error handling into the respective APIs `Download`, and `Pacakge`.
* ensure the Error raised has enough information to build up an appropriate response to the user.
* use a single `catch` rather than multiple catches between operations.

Not a blocker, but this is the approach I've been using.

ceejbot: If I understand this logic, you don't have one promise chain here, really. You actually have two. First action is to fetch the package. If that fails, then you have one set of `reply.view()` calls. If that succeeds, you have another promise chain (well, a very short chain) starting with `Download.getAll(package.name)` and ending with another set of `reply.view()` calls. This chain doesn't break on an error from the downloads service.

In this case I'd back up & change the api for `Download.getAll(package.name)`. You don't want this to ever resolve to an error because you are always going to swallow those errors in this consumer. Instead, you want it to return null or empty list or whatever it is that you use to signal the absence of downloads. This allows you to handle the real error case (failure to fetch a package) separately.

ceejbot: With this revision, the logic will look more like this:

```javascript
Package.get(p)
.then(function(p)
{
  return Download.getAll(p);
})
.then(function(dls)
{
   // do stuff, render page
})
.error(function(err)
{
  // handle error rendering & logging
});

zeke:
Closing. Replaced by https://github.com/npm/newww/pull/708

PR #687, opened 2015-02-27
It will not be yours

zeke:
What you see when you're logged in:

screen shot 2015-02-26 at 4 57 14 pm

What you see when you're not logged in:

screen shot 2015-02-26 at 4 57 41 pm

What you still see if the package is unscoped:

screen shot 2015-02-26 at 4 58 27 pm


zeke:
I wanna update this to use https://www.npmjs.com/package/npm-package-arg before shipping.

zeke:
Note: show "it will be yours" if the given scope === the logged in username

handlers/fallback.js
```diff
@@ -32,7 +32,7 @@ module.exports = function (request, reply) {
         if (validatePackageName(route).validForNewPackages) {
           opts.package = {name: route};
         }
-        return reply.view('errors/not-found', opts).code(404);
+        return reply.view('errors/package-not-found', opts).code(404);
```
rockbot: what about for routes that aren't packages? i.e. `/policies/blerg`?

We can leave it as packages, but I just want to bring up this pesky edge case :-)

PR #690, opened 2015-02-28
no mocks

zeke:
refactored handlers/models/tests, per the discussion with ceej and rockbot on monday. We are moving away from the "override the models with fake models" strategy in favor of the ben_jamincoe style of using nock mocks in both the model tests and the controller (AKA handler) tests. The main benefit of this change is that it's more of an end-to-end test of What Actually Happens when a request is made.

This PR is ready to merge. It takes care of everything except `User`. That will have to happen another day.


handlers/collaborator.js
```diff
@@ -1,12 +1,7 @@
 var collaborator = module.exports = {};
 
-var collaboratorClientFromRequest = function(request) {
-  var bearer = request.auth.credentials && request.auth.credentials.name
-  return new request.server.models.Collaborator({bearer: bearer});
-}
-
 collaborator.list = function (request, reply) {
-  var Collaborator = collaboratorClientFromRequest(request)
+  var Collaborator = require("../models/collaborator").new(request)
```
rockbot: Why the extra `require("../models/collaborator")` here and not assign it to a variable at the top, i.e. 

```js
var CollaboratorModel = require("../models/collaborator");

// ... stuff happens

var Collaborator = CollaboratorModel.new(request);
```
?

models/download.js
```diff
@@ -12,6 +12,11 @@ var Download = module.exports = function (opts) {
   return this;
 };
 
+Download.new = function(request) {
+  var bearer = request.auth.credentials && request.auth.credentials.name;
+  return new Download({bearer: bearer});
```
rockbot: I don't think we need the `bearer` token for Downloads...? (Probably not a big deal, but just thought I'd mention it.)

zeke:
I updated the Collaborator handler based on your feedback. What do you think of this? https://github.com/npm/newww/blob/no-mocks/handlers/collaborator.js

As for the downloads bearer thing, it's there for the day in the future when download counts are available for private packages.

PR #699, opened 2015-03-03
hide the package nav behind FEATURE_ACCESS env var

zeke:
This will allow us to ship r2-powered www without all the private package bells and whistles.


zeke:
Busted! Back to the test mines for me...

zeke:
Updated. Ready to merge.

PR #701, opened 2015-03-03
ajax downloads

zeke:
## What I Did

- borrowed the existing handlebars template from the package page
- dropped [nets](https://www.npmjs.com/package/nets) in place of `request` in the download client, because it shims `xhr` on the browser, but still uses `request` in node-land
- removed bearer token on ajax request, because it's not allowed and throws a browser error.
- did some wonky stuff to borrow handlebars helpers for client-side use
- added api.npmjs.org to the CSP
- added a timing function that `console.log`s the time it took to fetch download counts

## Results

It takes close to a second to grab counts for a package. Average on my machine on the npm wifi is about `900ms`.

## To Do

If this PR survives and becomes a thing, the follow work remains to be done:

- [ ] add the client-side stuff to the homepage too
- [ ] don't try to fetch counts for scoped packages
- [ ] figure out how to use no promises, or a smaller promises thing like `p-promise` because the current implementation is adding 100+ KB to the minified bundle
- [ ] remove download stuff from package handler
- [ ] remove download stuff from homepage handler


zeke:
I just updated the download model to use `promise` instead of `bluebird`. Tests still pass without any modifications, and the browser bundle size went down from `270K` down to `175K`. I also removed the `lodash` dep from the download client too in favor of a more basic "constructor". This shaved another `25K` off, and the bundle is back down to a cool `174K`. Feeling pretty good about that.

zeke:
Good learnin's in here, but not gonna shipping this. Instead: https://github.com/npm/newww/pull/708

PR #703, opened 2015-03-04
adds hubspot tracking for private npm signups

rockbot:


handlers/customer.js
```diff
@@ -20,6 +21,8 @@ billing.getBillingInfo = function (request, reply) {
 };
 
 billing.updateBillingInfo = function(request, reply) {
+  var sendToHubspot = request.server.methods.npme.sendData;
```
zeke: `sendData` is very vague. If it's only used for this purpose, maybe a rename is in order?

config.js
```diff
@@ -88,7 +88,8 @@ config.license = {
     "form_npme_signup": "12345",
     "form_npme_agreed_ula": "12345",
     "form_npme_contact_me": "12345",
-    "form_private_npm": "12345"
+    "form_private_npm": "12345",
+    "form_private_npm_signup": "12345"
```
zeke: Is there a reason to keep fake hubspot form IDs in here? They end up being exposed client-side anyway. I would prefer they all lived in `lib/hubspot.js` or something. That would get us one step closer to eradicating `config.js`, and would mean less futzing with the `newww-config` repo.

handlers/customer.js
rockbot: it's used in multiple files across the codebase for sending data to hubspot... agreed, though, that we can create a Hubspot model along the lines of the User/Package mdoels.

test/adapters/redis-sessions.js
```diff
@@ -37,6 +37,7 @@ describe('redis-requiring session stuff', function() {
   before(function (done) {
     client = redis.createClient(config.port, config.host);
     client.auth(config.password, function () {});
+    client.flushdb();
```
zeke: Why call `done` in the after block but not the before block?

handlers/customer.js
ceejbot: Shave that yak later: no need to rename an existing function right now.

config.js
ceejbot: Ideally the entire feature is disabled in non-npm deployments of the website, so the config should end up being irrelevant to non-npm developers. Eventually. There is no need to do that work right now. I'd say keep doing config as we have been, and deal with it as a lump when we decide what "dealing with it" looks like for real.

test/adapters/redis-sessions.js
rockbot: it is, just not inside `client.flushdb()` (see https://github.com/npm/newww/blob/hubspot-integration/test/adapters/redis-sessions.js#L38-L44)

zeke: But doesn't that mean `done` will potentially be called before the db is flushed?

rockbot: isn't flushdb() technically an atomic function? I don't know that it's slow enough to realistically cause a race condition.

ceejbot: IIRC redis will sequentially process commands from any single client connection, so this will complete before the next redis command is issued in the tests. If the tests are passing, merge and move on.

PR #705, opened 2015-03-05
new browse routes

zeke:
We used to be doing all the browse stuff with a single handler and a single template.

This update breaks each browse type into its own explicit route, handler, and template.

I also came up with this One Weird Trick to create a share pagination helper for all these new handlers.


zeke:
Ready to merge!

facets/registry/show-browse.js
```diff
@@ -1,93 +0,0 @@
...
-    });
-
-    opts.prevUrl = opts.prevPage && "/browse/" + opts.browseby + "?page=" + opts.prevPage;
-    opts.nextUrl = opts.nextPage && "/browse/" + opts.browseby + "?page=" + opts.nextPage;
-
-    opts.paginate = opts.prevPage || opts.nextPage;
-
-    return reply.view('registry/browse', opts);
-  });
-};
```
rockbot: I love that this entire file is disappearing. omg so great.

mmalecki:
@rockbot re 2), do you have all the packages indexed locally? If the indexed set isn't complete, the result set you're getting from depended-upon will be incomplete too.

bcoe:
@rockbot, also the packages are currently different on staging. we're indexing from a CouchDB that has scoped modules, so that we can test the new functionality -- we've got a production server up now, perhaps we could point at this for a bit and confirm that things look identical.

Issue #654, opened 2015-02-19
finish off the password/email change flow


ceejbot:
This was broken by the incomplete switch to the new email module. Verify that it works again post-switch.

PR #706, opened 2015-03-05
Fix /password and /email-edit

rockbot:
Addresses #654 :-)

### DIFF
--- a/facets/user/show-email-edit.js
+++ b/facets/user/show-email-edit.js
@@ -1,7 +1,9 @@
 
-var userValidate = require('npm-user-validate'),
+var _ = require('lodash'),
+    userValidate = require('npm-user-validate'),
     crypto = require('crypto'),
-    utils = require('../../lib/utils');
+    utils = require('../../lib/utils'),
+    UserModel = require('../../models/user');
 
 var from, host;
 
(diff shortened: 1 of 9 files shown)


facets/user/show-email-edit.js
```diff
@@ -37,9 +39,9 @@ module.exports = function (options) {
 
     if (request.method === 'post' || request.method === 'put') {
       var data = request.payload,
-          email2 = data.email;
+          emailTo = data.email;
 
-      if (!email2 || userValidate.email(email2)) {
+      if (!emailTo || userValidate.email(emailTo)) {
```
zeke: I like this change.

facets/user/show-email-edit.js
```diff
@@ -237,7 +238,7 @@ function revert (request, reply) {
       return;
     }
 
-    var email1 = cached.changeEmailFrom,
+    var emailFrom = cached.changeEmailFrom,
```
zeke: And this one.

test/models/user.js
```diff
@@ -66,7 +66,23 @@ describe("User", function(){
         done();
       });
     });
+  });
+
+  describe("verifyPassword", function () {
+    it("is essentially login with separated params", function (done) {
```
zeke: Why not just re-use the login method? Not suggesting it, just wondering why.

rockbot: I just thought it would be clearer for users... from a code-as-documentation perspective, I thought it would be fairly confusing to see `login` on password change/email change, though I can be convinced otherwise.

PR #708, opened 2015-03-06
redis cache thing

zeke:
@ceej I started adding more tests, but got stuck with sinon and scope stuff and need your input:

- Some of our tests were (wrongly) passing before because we had `to.be.true` assertions instead of `to.be.true()`
- The `get() calls _fingerprint()` test is only passing now because I changed the code to use `exports._fingerprint()`. This is obviously not the right way to do it. Please advise.
- I suspect this is the same reason the logger tests are failing too.
- In the later tests I'm calling `cache.redis.set.restore();`, but subsequent tests are whining: `Attempted to wrap set which is already wrapped`.. is this a race condition or something?


zeke:
see also: https://github.com/npm/newww/pull/709

PR #709, opened 2015-03-06
redis cache thing: an alternative

zeke:
This is a refactor of https://github.com/npm/newww/pull/708 that I find easier to reason about scope-wise than the one we worked on earlier today. I was able to get all the tests passing. @ceejbot take a look and tell me what you think. I'm happy to toss this implementation out and proceed with the old one if you don't like it.


zeke:
Got it. My goal with this refactor was not really to make multiple caches an option, so much as to write it in a way that the scope of all the internal stuff we were writing tests for could be easily grokked.

zeke:
No reason it can't be parallel.

Please give feedback on how to get tests passing on the original ceejcache and I'll work on it tonight.

PR #711, opened 2015-03-08
Browserify friendly

zeke:
- replace `bluebird` with `promise` which has a compatible API but a smaller browser footprint
- remove instances of `request.get`, `request.post`, etc  with `method: "GET"`. This paves the way for using `nets` or `httipfy` in place of `request`.


zeke:
Now is not the time.

PR #717, opened 2015-03-09
adds backend for stars

rockbot:
the logic is in place, but the user-acl endpoint doesn't exist yet.

issue for user-acl: https://github.com/npm/user-acl/issues/68


zeke:
I've been moving handlers to the `/handlers` directory as they are rewritten to not use couch, and also moving their tests to `/test/handlers`, the idea being that whatever remains in `facets` is the stuff that hasn't been addressed yet. Are you okay with this?

test/registry/star.js
```diff
@@ -14,6 +15,8 @@ var server,
 
 // prepare the server
 before(function (done) {
+  process.env.USER_API = "https://user-api-example.com";
```
zeke: user.js now defaults to `https://user-api-example.com` if the envionrment variable is not present, so this should not be necessary.

rockbot: for reasons I cannot understand, the tests wouldn't pass without it. halp? :-)

Issue #672, opened 2015-02-24
logout


rockbot:
login via models.User().login()... but where is logout?

PR #723, opened 2015-03-11
updates logout handler

rockbot:
Addresses #672

### DIFF
--- a/services/user/methods/sessions.js
+++ b/services/user/methods/sessions.js
@@ -36,32 +36,29 @@ module.exports = {
     return function (user, next) {
       var start = Date.now();
 
-      request.server.methods.user.logoutUser(user.token, function () {
+      if (!user.sid) {
+        request.auth.session.clear();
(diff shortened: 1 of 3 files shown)


services/user/methods/sessions.js
```diff
@@ -36,32 +36,29 @@ module.exports = {
...
 
-        if (!user.sid) {
-          request.auth.session.clear();
-          return next(null);
+      request.server.app.cache.drop(user.sid, function (err) {
+        if (err) {
+          request.logger.error(Boom.internal('there was an error clearing the cache'));
+          request.logger.error(err);
+          metrics.metric({name: 'delSessionError'});
+          return next(err);
```
ceejbot: Should this really bubble up to the user? Can we at all tidy up & hide our messes from them? They were just trying to log out & might not care about the cache.

rockbot: nope, good catch! fixed.

PR #725, opened 2015-03-11
Added react to the list.

hemanth:
It would be nice to have that in the list, right? :)

//cc @zeke @isaacs


zeke:
How did we miss this!?

hemanth:
Heh heh thanks! :+1:

PR #730, opened 2015-03-11
package access page

zeke:
- [x] add collaborators
- [x] delete collaborators
- [ ] get back a single collaborator instead of many. see https://github.com/npm/user-acl/issues/77
- [ ] toggle permissions
- [ ] username autocomplete
- [ ] set page title on access page
- [ ] add front-end testing
- [x] account for `.live()` behavior
- [ ] error adding collaborator to package '@zeke/ord': duplicate key value violates unique constraint "user_team_user_id_team_id_idx")
- [ ] toggle package privacy


zeke:
@rockbot I've added the feature flag enforcement and tests to this branch. Let's merge this to `r2` and I will reconstitute the remaining access page tasks in a new PR:

- [ ] account for collaborator API response changes, pending https://github.com/npm/user-acl/issues/77
- [ ] enable package privacy toggler, pending https://github.com/npm/user-acl/issues/79
- [ ] disable permission toggler on global and public-scoped packages, pending https://github.com/npm/user-acl/issues/79
- [ ] enable autocomplete on "add collaborator" form
- [ ] set page title on access page
- [ ] add front-end testing

assets/scripts/crumb.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function() {
+  $(function(){
+    try {
+      window.crumb = $("[data-crumb]").data().crumb
```
rockbot: this seems counter-intuitive to me? you're only using it twice, and it only needs to be on forms...?

assets/scripts/package-access.js
```diff
@@ -0,0 +1,78 @@
...
+  return obj
+}
+
+ var errorHandler = function(xhr, status, error) {
+   console.error(error)
+   $("p.error").text(error).show()
+ }
+
+ var successHandler = function(data) {
+   console.log(data)
```
rockbot: do we need to bubble this up?

handlers/collaborator.js
```diff
@@ -16,20 +17,20 @@ collaborator.list = function(request, reply) {
 
 collaborator.add = function(request, reply) {
   Collaborator.new(request)
-  .add(request.params.package, request.payload.collaborator, function(err, collaborator) {
+  .add(request.packageName, request.payload.collaborator, function(err, collaborator) {
```
rockbot: do you want to add any validation to the payload? whenever I see `request.payload` I get a little nervous... might be a good idea to bring Joi in :-)

models/collaborator.js
```diff
@@ -20,27 +20,27 @@ Collaborator.new = function(request) {
...
+    request(opts,
       function(err, resp, body) {
-        if (err) {
-          return reject(err);
-        }
+        if (err) {return reject(err);}
         if (resp.statusCode > 399) {
-          err = Error('error getting collaborators for package: ' + package);
+          err = Error(fmt("error getting collaborators for package '%s': %s)", package, resp.body));
           err.statusCode = resp.statusCode;
```
rockbot: it would be good to log these errors as well

models/download.js
```diff
@@ -12,11 +12,14 @@ var Download = module.exports = function (opts) {
...
-    bearer: bearer,
-    cache: require("../lib/cache")
-  });
+  var opts = {}
+  if (request && request.auth && request.auth.credentials) {
+    opts.bearer = request.auth.credentials.name
+  }
+  if (String(process.env.NODE_ENV).match(/(production|staging|dev)/)) {
+    opts.cache = require("../lib/cache")
+  }
```
rockbot: when do we not have a NODE_ENV that doesn't match production/staging/dev? and why would we not want caching involved in that case?

templates/layouts/default.hbs
```diff
@@ -30,7 +30,7 @@
   <![endif]-->
 
 </head>
-<body>
+<body {{#if crumb}}data-crumb="{{crumb}}"{{/if}}>
```
rockbot: again, I don't understand why we would have this here when it's already in every form for security reasons...?

test/adapters/bonbon.js
```diff
@@ -132,6 +132,7 @@ describe("bonbon", function() {
       expect(resp.statusCode).to.equal(200);
       expect(resp.headers['content-type']).to.match(/json/);
       expect(resp.result).to.be.an.object();
+      delete process.env.NODE_ENV
```
rockbot: why delete the NODE_ENV?

test/handlers/collaborator.js
```diff
@@ -81,7 +81,7 @@ describe('PUT /package/foo/collaborators', function () {
       expect(resp.result.collaborator.name).to.equal("wrigley_the_writer");
       done();
     });
-  });
+  }*/);
 
 });
```
rockbot: any particular reason why these tests are commented out? (just curious)

test/handlers/package.js
```diff
@@ -45,8 +47,8 @@ describe("package handler", function(){
       server.inject(options, function (response) {
         resp = response
         $ = cheerio.load(resp.result)
-        packageMock.done()
-        downloadsMock.done()
+        // packageMock.done()
+        // downloadsMock.done()
```
rockbot: do you want these commented out?

assets/scripts/crumb.js
zeke: Imagine this scenario:

1. Visit `/package/foo/access` and the server renders the whole page
1. Type in a new collaborator name and hit enter
1. An Ajax request is made and a new collaborator row is rendered client-side

The context object currently passed to the [collaborator template](https://github.com/npm/newww/blob/25fde64e3ca6faf7768c8b6b21df0ec874037f5b/templates/partials/collaborator.hbs] looks like this:

```js
{
  name: "bob",
  permissions: "write"
}
```

In order to make the crumb available here, we'd have to update the collaborator handler to return something like this instead:

```js
{
  crumb: "12345",
  collaborator: {
    name: "bob",
    permissions: "write"
  }
}
```

This doesn't seem so bad, but then it means we can't share the same template between server and client, because server is doing this:

```hbs
{{#each package.collaborators}}
  {{> collaborator}}
{{/each}}
```

Sticking the token in the DOM makes it possible for client-side rendered forms to work with the current templates. Maybe there's a more elegant way to solve this problem?

assets/scripts/package-access.js
zeke: Nope. I was just using it for debugging. Removed.

handlers/collaborator.js
zeke: user-acl is already validating input and it's an interface we own/trust, so I'm not sure it's necessary.

models/collaborator.js
zeke: these errors go to the collaborator handler, which does this:

```js
if (err) {
  request.logger.error(err);
  return reply(err)
}
```

Is that what you mean? Or are you saying they should be logged within the model à la [models/user.js](https://github.com/npm/newww/blob/25fde64e3ca6faf7768c8b6b21df0ec874037f5b/models/user.js#L15-L21)?

models/download.js
zeke: Yeah that's not a great workaround, and I'm not happy with it either. Here's why I added it:

Using the cache in tests leads to nondeterministic results. The first time you run the tests you'll get one code path, subsequent tests will hit another. And because the cache module is not instantiated, its configuration persists across tests. By the time the `test/cache.js` tests are run, the cache has already been configured by the mock server, so the "cache should complain about not yet being configured" tests fail. Enabling the cache in tests also means writing nock mocks that can't be `done()`ed, because the request is only made on cache misses.

templates/layouts/default.hbs
zeke: It's not present in client-side generated forms.

test/adapters/bonbon.js
zeke: So it doesn't persist to other tests.

test/handlers/collaborator.js
zeke: because they don't pass yet :P

test/handlers/package.js
zeke: nope. removing.

test/handlers/collaborator.js
zeke: They depend on user-acl changes that haven't been made yet.

zeke:
Reading about caching in Hapi at http://hapijs.com/tutorials/caching#server-methods

> server.method() created a new policy with segment: '#sum' automatically for us. Also the unique item id (cache key) was automatically generated from parameters. By default, it handles string,number and boolean parameters. For more complex parameters, you have to provide your own generateKey function to create unique ids based on the parameters - check out the server methods tutorial for more information.

If we want to use Hapi's `server.method` caching strategy for downloads, we'll need to implement this `generateKey` function. Fortunately, @ceej and I already did that work in [cache._fingerprint](https://github.com/npm/newww/blob/ec73d1b24c8d59684f6b7a5587b04ec8b4123050/lib/cache.js#L33-L49).

@rockbot let's pow-wow and decide how to move forward on this.

zeke:
On second though, the parameters to functions like `download.getSome()` are just strings, so maybe the built-in key generator will work. The thing I'm still confused about with this `server.method` approach is how to get the per-request bearer token into the mix...

handlers/package.js
```diff
@@ -5,12 +5,13 @@ var npa = require('npm-package-arg');
...
   var package;
+  var name = request.packageName;
   var context = {title: name};
   var loggedInUser = request.auth.credentials;
   var Package = require("../models/package").new(request)
-  var Download = require("../models/download").new(request)
-  var name = request.params.package ||
-    request.params.scope + "/" + request.params.project;
+  var Download = require("../models/download").new({
+    request: request, cache: require("../lib/cache")
```
ceejbot: Nice dependency injection.

models/collaborator.js
ceejbot: So long as they're logged at least once, seems good.

models/download.js
ceejbot: Note here to summarize in-air discussion: this configuration decision will be made in one place somewhere (likely in the server startup that parses config) and a variable named something like `USE_CACHE` will be populated. That will make code like this much clearer, and also ensure that the decision is made once, consistently.

zeke:
Updated download handler to check for `USE_CACHE`. See also https://github.com/npm/newww-config/pull/23

models/collaborator.js
rockbot: it comes down to where/when we log errors... I think for now this is fine, but for the most part I like to log the error as soon as it happens.

zeke:
re: https://github.com/npm/newww/pull/730#issuecomment-82659263, I'm seeing that issue too. I hear from @bcoe that @mmalecki is working on couch<->user-acl data sync.

re: https://github.com/npm/newww/pull/730#issuecomment-82660576 maybe you need to add `FEATURE_ACCESS` to your `.env`?

Issue #731, opened 2015-03-11
port npm enterprise payment flow to r2


rockbot:
same bat flow, different bat channel. also different bats.
:mouse:

PR #735, opened 2015-03-13
Enterprise payments

rockbot:
Completes #731


assets/scripts/buy-enterprise-license.js
```diff
@@ -0,0 +1,82 @@
...
+  if (typeof StripeCheckout === 'undefined') {
+    return
+  }
+
+  $(function () {
+
+    var amount, subType, quantity
+
+    var handler = StripeCheckout.configure({
+      key: $('input[type=hidden]').data('key'),
```
zeke: This is making an assumption that there are no other hidden inputs on the page. Can we scope it more tightly?

assets/scripts/buy-enterprise-license.js
```diff
@@ -0,0 +1,82 @@
...
+        token.amount = amount
+        token.subType = subType
+        token.quantity = quantity
+        token.customerId = $('#customer-id').val()
+        token.crumb = $('input[name="crumb"]').val()
+
+        $.ajax({
+          url: '/enterprise/buy-license',
+          beforeSend: function(xhr) {
+            xhr.setRequestHeader('X-CSRF-Token', $('input[name="crumb"]').val())
```
zeke: @rockbot sorry I keep asking this, but can you remind me why this needs to be set in addition to the header?

assets/scripts/buy-enterprise-license.js
```diff
@@ -0,0 +1,82 @@
...
+          amount = 2500
+          quantity = 1
+          break
+        case 2:
+          description = "One-year license for Starter Pack"
+          amount = 30000
+          quantity = 1
+          break
+        default:
+          return false
```
zeke: Maybe this should console.error or something?

zeke:
Wow that's a lot of code! Looks good, pending the `process.env.STRIPE_PUBLIC_KEY` updates @rockbot and I discussed in-air.

assets/scripts/buy-enterprise-license.js
rockbot: the console.error will bubble up to the user - do we want that to happen?

assets/scripts/buy-enterprise-license.js
rockbot: I honestly don't know the answer. I will try to figure it out.

assets/scripts/buy-enterprise-license.js
rockbot: yes, good call - will do.

assets/scripts/buy-enterprise-license.js
zeke: It's not ideal, but it's better than nothing happening at all, right?

assets/scripts/buy-enterprise-license.js
ceejbot: node will lower-case the headers on the way in, so it's not going to be the case. I note that the code paths are completely different, so the answer has to lie in the jquery docs.

neonstalwart: this is a shot in the dark but does the request go directly to node or is there something between (like nginx) as a reverse proxy? maybe the header is not being forwarded?

zeke:
sweet! Once the button turns green, let's :shipit:

PR #746, opened 2015-03-17
[WIP] attempt to cache download counts with hapi server methods

zeke:
Trying to make use of hapi's built-in catbox caching for fetching package download counts, instead of rolling our own. Download counts in this branch are currently not working when the cache is enabled.

### DIFF
--- a/handlers/package.js
+++ b/handlers/package.js
@@ -8,7 +8,7 @@ package.show = function(request, reply) {
   var context = {title: name};
   var loggedInUser = request.auth.credentials;
   var Package = require("../models/package").new(request)
-  var Download = require("../models/download").new(request)
+
   var name = request.params.package ||
     request.params.scope + "/" + request.params.project;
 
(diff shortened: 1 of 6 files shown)


services/downloads/index.js
```diff
@@ -1,39 +1,24 @@
...
-  server.method('downloads.getAllDownloads', require('./methods/getAllDownloads')(options.url), {
-    cache: {
-      staleTimeout: 1 * SECOND, // don't wait more than a second for fresh data
-      staleIn: 60 * 60 * SECOND, // refresh after an hour
-      segment: '##alldownloads'
-    }
-  });
+  server.method('downloads.getAll', function(packageName) {
+    return require('../../models/download').new().getAll(packageName);
+  }, opts);
```
hueniverse: The method returns a promise but you forgot to set `opt.callback` to `false`.

handlers/package.js
```diff
@@ -60,7 +60,7 @@ package.show = function(request, reply) {
         return promise.cancel();
       }
 
-      return Download.getAll(package.name)
+      return request.server.methods.downloads.getAll(package.name)
```
hueniverse: This will turn into a callback method once caching is active because the cache doesn't understand promises.

zeke:
Putting this on ice. We can revisit later if needed..

PR #750, opened 2015-03-18
Update star routes

rockbot:
all fixed!! this is ready for review.
cc: @zeke @ceejbot


mmalecki:
:star: stuff looks good :+1:

PR #753, opened 2015-03-18
adds resend confirmation email route

rockbot:
Please note that the link in the template is totally just a placeholder for proof of concept :-)


zeke:
This looks great! Such readable code.

PR #768, opened 2015-03-23
access page

zeke:
Picking up where [the previous PR](https://github.com/npm/newww/pull/730) left off:

- [x] account for collaborator API response changes, pending https://github.com/npm/user-acl/issues/77
- [x] enable package privacy toggler, pending https://github.com/npm/user-acl/issues/79
- [x] disable permission toggler on global and public-scoped packages, pending https://github.com/npm/user-acl/issues/79
- [ ] enable autocomplete on "add collaborator" form
- [x] set page title on access page
- [x] redirect to access page after following its login prompt
- [x] hide user immediately when deleted
- [ ] make it slightly difficult for user to remove self


zeke:
The access page is mostly complete and I think this PR is ready to go. There's a stuff in here like better enforcement of the access feature flag that needs to land in `r2` before we can ship `r2`.

I have created 


@rockbot please review when you have time.

assets/scripts/package-access.js
```diff
@@ -1,77 +1,101 @@
...
+    .unbind("change", updateCollaborator)
+    .bind("change", updateCollaborator)
+
+  $('.remove-collaborator')
+    .unbind("submit", removeCollaborator)
+    .bind("submit", removeCollaborator)
+
+  $('#package-access-toggle')
+    .unbind("change", togglePackageAccess)
+    .bind("change", togglePackageAccess)
```
rockbot: genuinely curious: why `.unbind` and then `.bind` the same thing?

handlers/collaborator.js
```diff
@@ -18,12 +19,22 @@ collaborator.list = function(request, reply) {
 collaborator.add = function(request, reply) {
   Collaborator.new(request)
   .add(request.packageName, request.payload.collaborator, function(err, collaborator) {
+
+    //
```
rockbot: ghost comment! :ghost:

handlers/collaborator.js
```diff
@@ -18,12 +19,22 @@ collaborator.list = function(request, reply) {
 collaborator.add = function(request, reply) {
   Collaborator.new(request)
   .add(request.packageName, request.payload.collaborator, function(err, collaborator) {
```
rockbot: the `collaborator` returned in this callback will conflict with the object `collaborator` defined on line 4.

I recommend changing the callback variable name to reduce confusion.

assets/scripts/package-access.js
zeke: to prevent binding the handler multiple times. :)

handlers/customer.js
```diff
@@ -2,6 +2,11 @@ var billing = module.exports = {};
 var Customer = require("../models/customer").new();
 
 billing.getBillingInfo = function (request, reply) {
+
+  if (!process.env.FEATURE_ACCESS) {
```
rockbot: if `process.env.FEATURE_ACCESS` is set to `"false"`, this still won't execute. So maybe `if (process.env.FEATURE_ACCESS !== "true")` instead?

rockbot: same for other similar checks throughout this PR.

rockbot: Actually... would it be easier to set this at the top level in the routes file? or in `bonbon`? Just thinking of ways to keep it a bit DRYer

handlers/package.js
```diff
@@ -72,10 +72,21 @@ package.show = function(request, reply) {
 
       package.isCollaboratedOnByUser = Boolean(process.env.FEATURE_ACCESS)
```
rockbot: if `process.env.FEATURE_ACCESS` is set to "false" this will execute.

models/collaborator.js
```diff
@@ -27,9 +27,10 @@ Collaborator.prototype.list = function(package, callback) {
       method: "GET",
       url: url,
       json: true,
-      headers: {bearer: _this.bearer}
     }
 
+    if (_this.bearer) opts.headers = {bearer: _this.bearer};
+
```
rockbot: :+1:

routes/authenticated.js
```diff
@@ -66,5 +66,30 @@ module.exports = [
...
+      "/package/{package}/collaborators/{username}",
+      "/package/{scope}/{project}/collaborators/{username}",
+    ],
+    method: "DELETE",
+    handler: require('../handlers/collaborator').del
+  },{
+    path: "/package/{scope}/{project}",
+    method: "POST",
+    handler: require('../handlers/package').update
   }
```
rockbot: could easily put in here something like

```js
if (process.env.FEATURE_ACCESS) {
  routes = routes.concat(arrayOfCollaboratorBits);
}
```

Hapi will automatically send routes that it doesn't think exist over to a 404 not-found page.

templates/layouts/default.hbs
```diff
@@ -60,7 +60,7 @@
         {{else}}
           <a href="/login"><img src="/static/images/wombat-avatar-small.png" id="user-avatar" /></a>
           <ul class="single">
-            <li><a href="/signup">sign up</a> or <a href="/login">log in</a></li>
+            <li><a href="/signup">sign up</a> or <a href="/login?done={{canonicalURL}}">log in</a></li>
```
rockbot: do we need this `?done=` bit here? Doesn't `/login` automatically take users to the homepage?

templates/package/access.hbs
```diff
@@ -1,24 +1,33 @@
...
+{{/with}}
+
 {{> package-header package}}
 
 <div class="tablet-width container">
 
   {{#unless user}}
     <p class="notice please-log-in">
-      To configure access to this package, please <a href="/login">log in</a>.
+      You may be an admin of this package, but you'll never
```
rockbot: "you'll never know" seems slightly antagonistic... maybe something like "Are you an admin of this package? Why don't you <a>login</a>?"

adapters/bonbon.js
```diff
@@ -69,6 +69,14 @@ exports.register = function(server, options, next) {
         break;
     }
 
+    // Add feature flags to view context
+    if (process.env.FEATURE_ACCESS) {
```
ceejbot: This is a very generic name for a feature flag. What feature is being controlled by this flag? Is it intended to be a general dictionary of features that can be toggled? With a generic name like that, I would expect it to be. But a few lines down you use it to flip `access` to true. And the usage appears to treat it like a boolean.

If this is coming from `etcd` config (which it probably should be) it can be a full object with fields named for the features being turned on & off.

If it's just a boolean to control whether the access page is visible, it needs a name that reflects its meaning.

handlers/collaborator.js
zeke: removed

handlers/collaborator.js
zeke: fixed

handlers/customer.js
zeke: explicitly checking for "true" now.

handlers/package.js
zeke: explicitly checking for "true" now.

templates/layouts/default.hbs
zeke: This makes it redirect to the page they were on when they clicked `login`.

adapters/bonbon.js
zeke: what name would you prefer, @ceejbot?

templates/package/access.hbs
zeke: updated

adapters/bonbon.js
zeke: My thinking: `FEATURE_THIS`, `FEATURE_THAT`. In this case, the feature is `ACCESS`, and the aspects of the site that relate to it. Open to alternatives.

ceejbot: I see--  "package access page" gets shortened to the sadly generic "access" and my confusion resulted. I'd still rather see this being pulled from a dict of flippable features, projecting forward to lots of features rolled out in code before being turned on, or being turned on for only a beta-testing group or whatever. This would be what I alluded to last week about how a general feature flipping setup is going to come in handy pretty soon.

Something to think about for next time.

zeke:
This is ready for review again. Going to set up more granular feature flags in a different branch after this gets merged. Too many PRs in flight.

routes/authenticated.js
zeke: will do

zeke: Open closer inspection, this will be a lot of work, because it will mean reloading the test server for every test that used to be modifying environment on a per-request basis. Do you mind if we punt on this change until I implement a more granular feature flag thing in a subsequent PR?

zeke:
:emoji:

PR #777, opened 2015-03-26
Issue#508 added github contributor count on the package page

yashprit:
@zeke Please review this and let me know, I still feel this file need more cleanup, I see lots of duplicate code especially `switch` logic and `api` call, but as you mentioned start with small, so I just added code to show contributor and added async to make parallel api call.


zeke:
@yashprit I've started a full refactor using octokat. Take a look here: https://github.com/npm/newww/pull/824

If you want to keep collaborating on it, let me know and I can outline how I think it should be built. :)

yashprit:
@zeke I am interested, so please outline what do you think, I see some code you have already committed #824

zeke:
@yashprit I updated the description with an outline of the various pieces. It's a start, but there's probably more. Let me know what you think..

PR #797, opened 2015-03-30
cache user object

rockbot:
Per a conversation in slack:

> for every user, we are going to have two cache bits that we care about: the user-specific cached data and the user-session-specific cached data

> the user-specific cached data will simply be a cached version of the user object

> the user-session-specific cached data will essentially be a source of "this user is logged in via this specific interface" (i.e. devices/browsers/whatever)

> so when a user logs in/out, we validate/invalidate the session key

> and when a user changes their data, we validate/invalidate the user key

> this will result in having two calls whenever we need to verify that a user is logged in and we need to see the user information, but those calls will be to redis (which is super duper fast), as opposed to couch (which is comparably slower than snails)


adapters/bonbon.js
```diff
@@ -24,17 +25,30 @@ exports.register = function(server, options, next) {
       start: Date.now(),
     };
 
-    if (request.method !== "post") {
-      return reply.continue();
+    if (request.auth && request.auth.credentials) {
+      UserModel.new(request).get(request.auth.credentials.name)
+      .then(function(loggedInUser) {
+        request.loggedInUser = loggedInUser;
```
zeke: :)

zeke:
I'm getting an error after clicking a change password email link. This flow works in r2.

https://localhost:15443/forgot/yeO4XkdmgV4K3FteY0Z8RqH564afkoEESTVEcelL

screen shot 2015-03-31 at 5 24 13 pm

```
2015-04-01T00:22:22.267Z  warn 1427847742265:Zekes-MacBook-Air.local:12065:i7xznk8c:10014: About to change password { name: 'zeke' }
2015-04-01T00:22:22.575Z  info 1427847742265:Zekes-MacBook-Air.local:12065:i7xznk8c:10014: 127.0.0.1 - 1427847742265:Zekes-MacBook-Air.local:12065:i7xznk8c:10014 [31/Mar/2015:17:22:22 -0700] "GET /forgot/yeO4XkdmgV4K3FteY0Z8RqH564afkoEESTVEcelL http/1.1"  - 308ms
Debug: internal, implementation, error 
    TypeError: Uncaught error: undefined is not a function
  at /Users/z/n/www/facets/user/show-forgot.js:80:12
  at tryCatcher (/Users/z/n/www/node_modules/bluebird/js/main/util.js:24:31)
  at Promise.successAdapter (/Users/z/n/www/node_modules/bluebird/js/main/nodeify.js:22:30)
  at Promise._settlePromiseAt (/Users/z/n/www/node_modules/bluebird/js/main/promise.js:528:21)
  at Promise._settlePromises (/Users/z/n/www/node_modules/bluebird/js/main/promise.js:646:14)
  at Async._drainQueue (/Users/z/n/www/node_modules/bluebird/js/main/async.js:79:16)
  at Async._drainQueues (/Users/z/n/www/node_modules/bluebird/js/main/async.js:89:10)
  at Async.drainQueues (/Users/z/n/www/node_modules/bluebird/js/main/async.js:14:14)
  at process._tickDomainCallback (node.js:376:13)
```

facets/company/show-static.js
```diff
@@ -1,7 +1,6 @@
 module.exports = function (type, text) {
   return function (request, reply) {
-    var opts = { user: request.auth.credentials },
-        timer = { start: Date.now() };
```
ceejbot: Well, that dead code had been lingering for a while! Happy to see it go.

models/user.js
```diff
@@ -83,10 +84,25 @@ User.prototype.login = function(loginInfo, callback) {
...
+  };
+};
+
+User.prototype.drop = function drop (name, callback) {
+
+  if (process.env.USE_CACHE === 'true') {
+    return cache.drop(this.generateUserACLOptions(name), callback);
+  }
+
+  return callback(null);
```
ceejbot: Zalgo has been released! Fire the callback on `process.nextTick()`or be very, very sure that no calling code ever accidentally depends on this call being async & the callback being fired after i/o.

models/user.js
```diff
@@ -83,10 +84,25 @@ User.prototype.login = function(loginInfo, callback) {
...
+    url: fmt("%s/user/%s", this.host, name),
+    json: true,
+  };
+};
+
+User.prototype.drop = function drop (name, callback) {
+
+  if (process.env.USE_CACHE === 'true') {
+    return cache.drop(this.generateUserACLOptions(name), callback);
+  }
```
ceejbot: You can remove a lot of repetitive & unimportant error logging in the calling layers if you log errors here instead of every single time you call `User.drop()`. Log the error here, always callback with no params, and clean up everything. You'll note that you never do anything other than log at the call sites (what can you do anyway?) so the error is something the lower layer should handle for you.

lib/cache.js
```diff
@@ -30,6 +30,19 @@ exports.configure = function configure(opts)
...
+{
+    if (redis) {
+        redis.quit();
+        exports.redis = redis = null;
+    }
+
+    options = null;
+    DEFAULT_TTL = 300; // seconds
+    KEY_PREFIX = 'cache:';
+    return callback();
```
ceejbot: I'm not sure why this function takes a callback if all of its work is synchronous.

lib/cache.js
```diff
@@ -54,7 +67,23 @@ function safeparse(input)
...
+{
+    assert(_.isObject(redis), 'you must configure the redis client before using the cache.');
+    assert(_.isObject(opts), 'you must pass a Request-ready options object to cache.get()');
+
+    var key = exports._fingerprint(opts);
+    redis.del(key, function(err)
+    {
+        if (err)
+        {
+            exports.logger.error('problem getting ' + key + ' from redis @ ' + options.redis);
```
ceejbot: My copy-pasta error here: this should probably log "problem dropping" instead of "getting".

models/user.js
ceejbot: And I just re-read the implementation of `cache.drop`-- it swallows all errors & logs already, so you both don't have to bother here AND you can delete all the spots where you log user cache drop errors in the callers!

models/user.js
```diff
@@ -94,15 +110,19 @@ User.prototype.get = function(name, options, callback) {
...
-      if (err) { return reject(err); }
-      if (resp.statusCode > 399) {
-        err = Error('error getting user ' + name);
-        err.statusCode = resp.statusCode;
-        return reject(err);
-      }
-      return resolve(body);
-    });
+
+    if (process.env.USE_CACHE) {
```
ceejbot: We discussed this briefly: this config option is best consulted & respected in the caching library itself, so the details of the implementation are hidden away from callers. This will also allow you to remove a lot of repeated code from every spot you use the cache.

PR #799, opened 2015-03-31
Fix url building for current url of paginated results.

re5et:
This was leading me to some very goofy search results.


zeke:
thanks!

PR #800, opened 2015-04-01
Refactor routes with DRY mindset

ericdouglas:


zeke:
Once upon a time this array style for HTTP verbs was broken in hapi, so we had to declare each verb in its own route. But maybe that's fixed in hapi now?

Also, we're just about to merge `r2` to `master`. It has a completely rewritten routes, so your patch would likely be unmergeable. Would you mind submitting these improvements to the `r2` branch instead?

ericdouglas:
@zeke thanks for the feedback! I will search for this problem with the *array style*, and if it make sense, of course I could send a new PR!

All the best!

npm <3

PR #804, opened 2015-04-02
remove unused dependencies

zeke:
I noticed a few unused dependencies introduced in a recent PR and wondered if there's a tool to track those down. Indeed there is: https://www.npmjs.com/package/depcheck

```sh
npm i -g depcheck && depcheck .

Unused Dependencies
* accept-language-parser
* alphabet
* async-cache
* cargo
* catbox
* chalk
* deck
* highlight.js
* hiredis
* nodemailer
* npm-email-templates
* npm-typeahead
* pkgs
* pretty-hrtime
* semver
* similarity
* tough-cookie

Unused devDependencies
* grunt
```

- I didn't remove `npm-email-templates` because it is being used unconventionally with `require.resolve`.
- I didn't remove `accept-language-parser` because I hope to use it for localization purposes someday.


othiym23:
`depcheck` is useful! One caveat: it doesn't deal well with packages deduped up the tree that are used further down the tree. I ran it against `npm` and was able to verify without much trouble that all of the indicated modules were used by other dependencies. This probably doesn't augur well for its ability to deal with `npm@3`, either.

zeke:
Good to know, @othiym23. I recognize most everything in the list above as stuff that we used at one time and forgot to yank out when we stopped using it. Tests are passing and a smoke test in the browser says things are working, but I'll leave it to @rockbot to confirm or deny whether we still need any of these deps.

othiym23:
My comment was more of a, "keep this in mind for the future" than a, "omg the universe will be as unto a smoking ruin should you use this!" It's a neat tool! And it looks like it works great (for now) for packages that do not have npm's particular peculiarities.

bcoe:
@zeke I don't know what to tell you about some of these dependencies but:

* my development environment would not install `newww` without `semver`.
* and travis was not building without `tough-cookie `. on a branch I was on.

It seems like something strange is happening when installing `newww` from a fresh install.

zeke:
@bcoe that is strange. This PR is passing on travis...

`¯\_(ツ)_/¯`

PR #807, opened 2015-04-02
use percentages for a more adaptive mobile/desktop package page

zeke:
This fixes that whitespace above the sidebar on package pages, and lowers the "switch to mobile" threshold from 900px to 767px.


zeke:
screen shot 2015-04-02 at 4 07 23 pm

screen shot 2015-04-02 at 4 07 31 pm

PR #821, opened 2015-04-06
no null bearer

zeke:
Don't add bearer token to request if user is not logged in.

This was crashing the profile page for logged-out users.


bcoe:
:+1: nice quick fix, great catch.

PR #764, opened 2015-03-22
hyperclient

zeke:
Hyperclient is an experimental abstraction of our models AKA clients AKA facades AKA the things that talk to our new ACL service. The goal is to reduce the amount of redundant code in `package`, `user`, `collaborator`, `customer`, etc. Things like:

- default logger
- bearer header
- `json` option for `request`
- fully qualifying relative URLs
- promise-wrapped requests

It uses ES6 stuff, so iojs (or maybe node 0.12?) is required. New thingies include:

- `let`
- `const`
- template strings
- `class`
- `extends`
- built-in promises

To run the tests:

```
npm run hypertest
```

I couldn't figure out how to pass harmony flags to `lab`, so I adapted the tests to use mocha, which has [documented iojs flags](https://github.com/mochajs/mocha/blob/2c2ed43ae7f047f333468e2043001bfb8c9d13d5/bin/_mocha#L81-L88
):

```
--harmony: enable all harmony features (except typeof)
--es_staging: enable all staged features
--harmony-collections: enable harmony collections (sets, maps, and weak maps)
--harmony-generators: enable harmony generators
--harmony-proxies: enable harmony proxies
--harmony_shipping: enable all shipped harmony fetaures (iojs)
--harmony_arrow_functions: enable "harmony arrow functions" (iojs)
--harmony_proxies: enable "harmony proxies" (iojs)
```

PR #825, opened 2015-04-07
Registry relational queries are now cached.

ceejbot:
The package "model" now uses the cache for all package data gets. It drops any cached data for a specific package on update. There is now a convenience function for generating the predictable package Request options, so it is guaranteed to look the same every time.

Added promisified versions of `cache.get()` and `cache.drop()` for convenience, called `cache.getP()` and `cache.dropP()`. Added unit tests for the promisified wrappers.

Adjusted two package "model" tests to look for the regularized error messages that the cache generates on 404s.

While I was looking at it, parallelized the queries that build the home page, which should result in a modest speed improvement.

All tests pass. This of course means success is assured.

cc @rockbot, @bcoe

### DIFF
--- a/handlers/homepage.js
+++ b/handlers/homepage.js
@@ -1,28 +1,30 @@
+var P = require('bluebird');
+
 module.exports = function (request, reply) {
   var Package = require("../models/package").new(request)
   var Download = require("../models/download").new({
     request: request, cache: require("../lib/cache")
-  })
+  });
   var context = {
     explicit: require("npm-explicit-installs")
-  }
+  };
 
-  Package.list({sort: "modified"})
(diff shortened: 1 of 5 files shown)


zeke:
[props](https://github.com/petkaantonov/bluebird/blob/master/API.md#props---promise) to the rescue. Thank you @ceejbot!

I "promise" to extract the common http client boilerplate into a reusable thing eventually, so things like `generatePackageOpts` won't need to bothered with. The [ES2015 hyperclient](https://github.com/npm/newww/pull/764) may be premature since we're leery of iojs' volatility, so an ES5 equivalent is probably in order...

PR #834, opened 2015-04-07
enables list-specific TTLs

rockbot:
@ceejbot @zeke what do you think about this approach? if it's kosher, I'll go ahead and update other spots where we're using `Package.list()`.

### DIFF
--- a/test/models/package.js
+++ b/test/models/package.js
@@ -157,8 +157,8 @@ describe("Package", function(){
 
       Package.get('@zeke/ord')
         .then(function(package) {
-          mock.done()
-          done()
+          mock.done();
+          done();
         });
     });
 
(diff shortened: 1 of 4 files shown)


zeke:
lovely

test/models/package.js
```diff
@@ -261,6 +261,78 @@ describe("Package", function(){
...
+          callback(null);
+        });
+
+        var options = {
+          author: "zeke"
+        };
+
+        Package.list(options)
+        .then(function(){
+          // mock.done();
```
zeke: is there a story here?

rockbot: what do you mean?

zeke: just wondering about the commented out mock.done().. I've had a few of these myself while fighting nock+lab. Wondering if you encountered the same, or if this line is just a relic.

rockbot: oh! yep, it's a relic - i shall remove it :-D

(you'll note there is no mock above it, to which it would have once referred ;-))

PR #841, opened 2015-04-08
banner

rockbot:
Here's how it works:

* It shows up on every page
* If a user clicks the "no thanks" button, we save a 'disable-banner' value in localStorage, and never show it again

Still needs styling but the functionality is in place.


zeke:
I think the banner should default to a hidden state, then reveal if appropriate. This will prevent the banner from flashing on the screen and disappearing when js loads.

zeke:
I also think the localStorage property name could be more specific to be more future-proof. Maybe something like `disable-private-modules-banner`?

zeke:
To reset:

```
localStorage.removeItem('disable-private-modules-banner')
```

zeke:
At last, a design that I don't think sucks:

screen shot 2015-04-13 at 1 20 56 pm

What do others think?

zeke:
Here's the behavior: 

- banner defaults to an invisible state
- banner is shown if the following criteria are met:
  1. date is after 6am pacific Tuesday (currently Monday for testing)
  1. notification hasn't been disabled
  1. page is anything but `/private-modules`

Clicking **any link** in the banner disables the banner on future pageviews.

To delete the localstorage disabled-ness: `localStorage.removeItem('disable-private-modules-banner')`

zeke:
@rockbot clicking the signup link should not disable the banner. Only links within the `#notification-banner` disable it.

zeke:
in-air compromise: continue disabling banner on all link clicks; improve profile-page sidebar with link to update billing info.

assets/scripts/banner.js
```diff
@@ -0,0 +1,23 @@
+module.exports = function() {
+  $(function(){
+    var ready = new Date() > new Date("2015-04-13T06:00:00-07:00")
+    var disabled = localStorage.getItem('disable-private-modules-banner')
+
+    if (!ready || disabled || location.pathname === "/private-modules") {
+      console.log("no banner")
```
rockbot: do we want to keep this `console.log`?

zeke: nerp

Issue #827, opened 2015-04-07
show collaborators panel on package page when viewed by sole collaborator


zeke:
The collaborators panel on the package page currently only shows up if there's more than one collaborator. This is fine because it reduces redundant UI.

If however, you're the owner of the package and sole collaborator, you no longer have a way to get to the access page.

PR #846, opened 2015-04-09
show collaborators sidebar even if there’s only one collaborator

zeke:
Fixes https://github.com/npm/newww/issues/827


bcoe:
@zeke awesome thank you.

PR #851, opened 2015-04-09
use new github-url-to-object with support for git+https URLs

### LINKED ISSUE #848: newest `npm-package-arg` output requires post-processing for GitHub URLs
The version of `newww` currently in produciton is using a version of `npm-package-arg` that produces repo URLs that require some post-processing:

`git+https:` sadness

The `git+https:` is the same both for display and for clicking, and unfortunately, it doesn't work so good for clicking. What you want is `https://github.com/npm/hosted-git-info`, or perhaps `https://github.com/npm/hosted-git-info#readme`, which is what `hosted-git-info`, which `npm-package-arg` uses to do URL normalization, returns from its `docs()` method. @iarna or I can help you figure out how to get `npm-package-arg` to give you what you need.

zeke:
Fixes #848

https://github.com/zeke/github-url-to-object/releases/tag/v1.5.0

Before and after:

screen shot 2015-04-09 at 11 37 13 am


othiym23:
LGTM 👍🏾

Issue #857, opened 2015-04-09
billing page error messaging


bcoe:
### observed

when the billing-api returns an error code, this does not seem to be reflected in the logs nor is the error message.

### expected

I don't think we should communicate all errors to the user, but it would be good if we logged this information for forensic debugging.

rockbot:
moving discussion over to #859

PR #859, opened 2015-04-10
tiny addition to logging

rockbot:
Attempts to address #857

Note: the [customer model](https://github.com/npm/newww/blob/master/models/customer.js) takes whatever errors are coming from LICENSE_API and spits them out directly to the request.logger via the [customer handler](https://github.com/npm/newww/blob/master/handlers/customer.js). So if we're not logging the LICENSE_API errors, it would seem they're not getting sent?

cc: @bcoe


zeke:
I'm assuming this is replaced by https://github.com/npm/newww/pull/862 ?

Issue #860, opened 2015-04-10
Add logging for all external API requests


rockbot:
> wherever we're interacting with an external API, I'd love to see all requests + status codes and the error message returned :smile:

zeke:
Two cents.

While I agree in principle that every request should be easily logged, I would rather we add logging only where we're lacking visibility or investigating bugs for the moment (I.e licensing API stuff) for the time being.

Later we can log all the things when we have something like hyperclient that wraps all requests. We have a lot of duplicate code debt to pay down in www already and I don't wish to prolong the phenomenon.

ceejbot:
Adding logging for interesting events is not adding technical debt. It is exactly the opposite. It will pay off in debuggability now, particularly since we have some interesting interactions with other services.

PR #862, opened 2015-04-10
adds logging for external apis, plus test fixes

rockbot:
addresses https://github.com/npm/newww/issues/860

note that we're now logging the `body` coming from the Request calls in errors - this is because the error messages that come from the LICENSE_API are actually being sent in the body, and not in the `err` param.

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -2,6 +2,7 @@ var
     _       = require('lodash'),
     assert  = require('assert'),
     bole    = require('bole'),
+    clf     = require('../lib/utils').toCommonLogFormat,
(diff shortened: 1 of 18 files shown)


lib/cache.js
```diff
@@ -2,6 +2,7 @@ var
     _       = require('lodash'),
     assert  = require('assert'),
     bole    = require('bole'),
+    clf     = require('../lib/utils').toCommonLogFormat,
```
ceejbot: This is right on the edge of being too short a name for me, but it is used consistently, so I shall chill.

PR #871, opened 2015-04-13
address broken IP address (fixes #858)

### LINKED ISSUE #858: internal ip addresses being sent to hubspot instead of real user ips
Rod reports 

> Something is going wrong with the way that IP addresses are captured when people submit a Hubspot form. Hubspot captures an IP address, but the IP address does not seem to be unique to the user. Maybe it's capturing something else, such as an IP address from our servers or from Fastly ... I am not sure ... just guessing. [...]
> Several different folks have the same IP address - 54.68.3.141 - which maps back to Amazon.
> It would be great if we can fix this for the info that feeds into the Hubspot drip campaigns as close to launch as possible.

This bug might exist in the web site or in the registry side or both, so I'm ccing @rockbot & @mmalecki to take a peek.

rockbot:
I would like to push this up to staging before we merge it in :-)

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -32,4 +32,8 @@ utils.toCommonLogFormat = function toCommonLogFormat (resp) {
   return [clientIp, clientId, userid, time, requestLine, statusCode, objectSize].join(' ');
 };
 
+utils.getUserIP = function (request) {
+  return request.headers['x-forwarded-to'] || request.info.remoteAddress;
+}
+
 module.exports = utils;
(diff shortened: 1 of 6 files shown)


zeke:
@rockbot looks good to me. feel free to blow away staging once you're done previewing the banner stuff.

lib/utils.js
```diff
@@ -32,4 +32,8 @@ utils.toCommonLogFormat = function toCommonLogFormat (resp) {
   return [clientIp, clientId, userid, time, requestLine, statusCode, objectSize].join(' ');
 };
 
+utils.getUserIP = function (request) {
```
ceejbot: +1 to wrapping this up in a util function

Issue #761, opened 2015-03-20
simplify community participation


zeke:
We are in the process of replacing the backend used by www.npmjs.com and the npm CLI. In the near future, the website will no longer talk to CouchDB, in favor of a new postgres-powered relational follower of the main CouchDB registry. There are many reasons for this transition, scalability and access to meaningful data among them.

All significant feature development for the past few months has taken place on the [r2](https://github.com/npm/newww/tree/r2) branch, which we expect to ship soon, on the order of days or weeks.

Because this new registry2 architecture uses lots of separate microservices, it can be cumbersome to configure a development environment. To simplify development setup, the `r2` branch no longer uses the vagrant image, but instead points to various services in a remote staging environment. As of this writing, that environment is only accessible to npm employees, but we will soon offer an alternative that allows the public to more easily bootstrap a newww development environment.

Our goal is that anyone should be able to easily set up a full working development version of npmjs.com. We look forward to a day in the near future where `npm install && npm run dev` is all it takes.

Also noteworthy: The test suite in the `r2` branch no longer hits the network or relies on any external services, so anyone should be able to check out that branch today and `npm install && npm test` without any special configuration.

eiriksm:
Sounds awesome, nice work!

zeke:
The `r2` branch was merged into `master` last week, so the above references to `r2` are now relevant to `master`:

- `master` branch no longer uses the vagrant image
- `master` branch test suite no longer hits the network

We don't yet have an answer to this, but we know it's important. Stay tuned.

seldo:
Plan of record: we are going to set up a set of small, disposable servers with sanitized/dummy registry data that anyone can use to develop the registry.

sobering:
Hey guys, I was just wondering if there was an ETA on this at all? I'm eager to contribute to the site!

seldo:
We need to hire some more engineers for www to get it done. (You can apply at npmjs.com/jobs if you'd like to contribute! ;-))

TheAlphaNerd:
This sounds like a fun problem to help out with :P

tiandavis:
Is it possible for non-employees to run the development server yet?

Reason I ask, is I was able to get through the startup steps, but the webapp doesn't load.

```
# run redis in a background process
redis-server&

# copy environment-based config/secrets
cp .env.example .env

# install deps
npm install

# run the hapi server
npm run dev
```

meleyal:
Any updates on this?

rockbot:
[...]

PR #882, opened 2015-04-14
Remove the VM and update README

zeke:
Related: https://github.com/npm/newww/issues/761

This removes the `/dev` directory and all references to the old VM setup in the readme, plus other post-vm cleanup.


seldo:
We still need to either provide people with a VM that runs redis or include instructions on installing it locally, as well as a way to get the other microservices, but that's all adequately captured by https://github.com/npm/newww/issues/761

Issue #885, opened 2015-04-14
Errors with Billing Page


rosskukulinski:
I'm trying to submit my payment information through your website, but am getting console/network errors on OSX 10.10.3 with Chrome 41.0.2272.118 (64bit).  

When I load this page: https://www.npmjs.com/settings/billing, I get the following error in my browser console:
```
Refused to frame 'https://js.stripe.com/v2/channel.html?stripe_xdm_e=https%3A%2F%2Fwww.npmjs.com&stripe_xdm_c=default955599&stripe_xdm_p=1#__stripe_transport__' because it violates the following Content Security Policy directive: "frame-src https://checkout.stripe.com".
```

I'm also seeing a POST/415 error to csplog in the attached screenshot:
![screen shot 2015-04-14 at 1 04 42 pm](https://cloud.githubusercontent.com/assets/2746479/7146214/dbea5ace-e2a6-11e4-9ae5-00bfd83f0a01.png)

rosskukulinski:
Background info: I'm trying to add private repos for https://www.npmjs.com/~yodlr, but am unable to publish a private repo even after submitting billing details (Yes, I've ```npm login``` again and also cleared .npmrc and logged in again).

From the docs, I _think_ I should be publishing package.jsons with :
```
  "name": "@yodlr/mypackagename",
```

zeke:
Hey @rosskukulinski. Thanks for reporting. We had some bugs in our Content Security Policy that were affecting some users last week. You should be good to go now.

PR #886, opened 2015-04-14
Update Stripe config in CSP

zeke:
This is a fix for https://github.com/npm/newww/issues/885

Not sure why we weren't seeing this until now, but the stripe code currently expects to be able to use an iframe. 

When adding stripe to the `frameSrc` whitelist, I noticed that [frame-src is deprecated](https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#child-src), so I updated the CSP to use `child-src` instead.

This is up on staging now for testing.

### DIFF
--- a/lib/csp.js
+++ b/lib/csp.js
@@ -43,7 +43,8 @@ csp.default = {
   // objectSrc: Values for the object-src directive,
   // mediaSrc: Values for the media-src directive,
   frameSrc: [
-    'https://checkout.stripe.com'
+    'https://checkout.stripe.com',
+    'https://js.stripe.com',
   ],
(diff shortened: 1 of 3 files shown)


zeke:
Irony: MDN says to use `child-src`, but Firefox doesn't like it.

- Chrome, unpaid user: no error
- Chrome, paid user: no error
- Firefox 34, unpaid user: `Content Security Policy: Couldn't process unknown directive 'child-src'`
- Firefox 34, paid user: `Content Security Policy: Couldn't process unknown directive 'child-src'`

zeke:
Updated to Firefox 37: same result

PR #893, opened 2015-04-16
Probably not fixing the memory leak.

ceejbot:
Code inspection of the massive changes in the R2 branch did turn up some areas for cleanup: 
- don't make any more than one mailchimp client, since the config is invariant
- don't recalculate feature settings, since they're also invariant


zeke:
:shipit:

PR #911, opened 2015-04-22
add Twitter tracking codes

linclark:
This adds twitter tracking codes so we can track conversions based on tweets.

There is one on the profile page. This should only be used if the person has completed their purchase, and should only be shown once, but I couldn't figure out where the thank you message was generated.

### DIFF
--- a/templates/user/billing.hbs
+++ b/templates/user/billing.hbs
@@ -137,3 +137,10 @@
 </div>
 
 <script type="text/javascript" src="https://js.stripe.com/v2/"></script>
+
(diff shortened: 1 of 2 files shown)


zeke:
These two snippets look similar enough that it might make sense to abstract them into a common partial that takes args. To add a new a partial, add a file to `templates/partials/` and its filename will become the partial name.

Something like this should work:


```hbs
{{> twitter-tracking "l5xz2"}}
```

zeke:
Picking it up at https://github.com/npm/newww/pull/913

PR #913, opened 2015-04-23
twitter tracking

zeke:
A followup to https://github.com/npm/newww/pull/911

### DIFF
--- a/assets/scripts/twitter-tracking.js
+++ b/assets/scripts/twitter-tracking.js
@@ -1,0 +1,11 @@
+$(function(){
+  if (typeof twttr === "function") {
+    var pid = $("[data-twitter-pid]").data('twitterPid');
+    if (pid) {
+      twttr.conversion.trackPid(pid, {
+        tw_sale_amount: 0,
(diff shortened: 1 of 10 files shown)


zeke:
@linclark this is ready for review. It's up on [staging](https://newww-r2-1-west.internal.npmjs.com/~). Is there a way to test that it's working end-to-end from staging? If not, I'm comfortable shipping it to production. As an end user, all you need to do is drop this into any template:

```hbs
{{> twitter-tracking pid='l5xz2'}}
```

assets/scripts/twitter-tracking.js
```diff
@@ -0,0 +1,11 @@
+$(function(){
+  if (typeof twttr === "function") {
+    var pid = $("[data-twitter-pid]").data('twitterPid');
+    if (pid) {
+      twttr.conversion.trackPid(pid, {
+        tw_sale_amount: 0,
+        tw_order_quantity: 0
+      });
```
linclark: it looks like these aren't placed in the template.

PR #914, opened 2015-04-23
Disallow private toggle for unpaid users

zeke:
It is currently possible for an unpaid user to toggle their public package to private, effectively locking themselves out of the package until they become a paying customer. This PR addresses that problem.

When all of the following criteria are met...

- package is scoped
- package is public
- user is logged in
- user has write permissions to package
- user is not a paying customer

...then the public/private toggle is hidden and replaced with a notice:

screen shot 2015-04-22 at 8 59 37 pm

If the user is a paying customer, they see the private/public toggler:

screen shot 2015-04-22 at 9 00 02 pm


seldo:
This looks like it works, but I thought the plan was to add an isPaidUser() attribute to the user that we populate at login time?

PR #917, opened 2015-04-23
move twitter signup tracking code

zeke:
It was on the profile page. Now it's on the billing page. So the following actions will presumably satisfy the conversion requirements:

1. sign up (as an npm user)
1. confirm email
1. go to billing page

### DIFF
--- a/templates/user/billing.hbs
+++ b/templates/user/billing.hbs
@@ -18,7 +18,7 @@
   {{/if}}
 
   {{#if updated}}
-    {{> twitter-tracking pid='l5xz2'}}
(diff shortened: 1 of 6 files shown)


zeke:
From @seldo

> The conversion goal called "private modules purchase" is `l5xyy`; it should be on the "you entered your credit card, thank you" page

> The conversion goal called "private modules billing signup page" is `l5xz2`, it should be on the form where we ask for your credit card

> And the conversion goal I mentioned, where you simply create an account, would be on the email verification page, or some other page you only see while in the process of signing up, not just editing your profile or whatever.

seldo:
LGTM and yay for tests

Issue #916, opened 2015-04-23
cache is-paid status in the user object


ceejbot:
In the [handler for the access page](https://github.com/npm/newww/blob/c92f1dcf20f84b30e737aa5ffdb136a9eafcfb94/handlers/access.js#L60) we're making a call to the user-acl to snag the customer record. This is cache-able data since we'll want to use that all over the place. In fact, we should probably cache the whole user object, including stars + packages + customer status. 

Since we're now making several requests to build the user object, catbox is the right place to put this & not the Request-object cache. Modify the `dropCache` in the `User` model to drop the catbox version instead of the Request-object cache version.

ceejbot:
This remains high priority.

PR #921, opened 2015-04-24
Cache logged-in user data, with customer status mixed in

zeke:
An attempt to address https://github.com/npm/newww/issues/916

This breaks 50+ tests, but it appears to work per my smoke tests. @ceejbot please give it a once-over and let me know if anything looks way off.


zeke:
@ceej's above comments make perfect sense to me, but when it came time to implement, I got very confused and befuddled by the state of the code. I took a two-hour stab at refactoring it into something usable, but it just got more and more unwieldy and confusing. Requiring that the "models" know something about the `request` object or the `server` to which they are attached makes it hard to test those things in isolation, especially given that all of our model tests make no assumption about a server. Busting the catbox cache is also difficult, as the cache invalidation arguments must be the same as the arguments used to cache the request in the first place, so `request` cannot be passed to the server method, because every request is unique.

TLDR: I tried to refactor this but it's beyond my capacity to grok. 

Following conversations yesterday with @bcoe and @seldo, I'm going to bite the bullet and get the `acl-client` into a working state. I've ported our hand-rolled redis cache into `user-acl` and should have a PR ready that satisfies the `is-paid` requirement by the end of today.

zeke:
Made a lot of progress today. This is not ready for merge, but there's lots to see:

- [X] Add caching support to [acl-client](https://www.npmjs.com/package/@zeke/acl-client), mostly a straight copy of the ceejcache.
- [X] Move request object fingerprinting into a standalone module: [request-object-fingerprint](https://www.npmjs.com/package/request-object-fingerprint/)
- [X] Use acl-client to [fetch and cache logged-in user and customer data](https://github.com/npm/newww/blob/is-paid/adapters/bonbon.js#L39-L47) from two services
- [X] Start using [scoped packages in newww](https://github.com/npm/newww/blob/cbab373cf55891f2404b2c471932a37ea5b52842/package.json#L27). Yay dog-fooding!
- [X] Add [customer API endpoints](https://github.com/npm/acl-client/commit/92666f749fbd06c72f1b6b0c0326d45874f6f06d) to the schema. This means the client is talking to multiple services!
- [X] Keep internal [hostnames out of the code](https://www.npmjs.com/package/@zeke/acl-client#environment-variables), so package can be made public.
- [X] Add [everyone at npm](https://www.npmjs.com/package/@zeke/acl-client/access) as a collaborator
- [X] Infer `logger` from hapi request
- [X] Infer `bearer` from hapi request
- [X] Avoid hitting redis for [static asset requests](https://github.com/npm/newww/blob/cbab373cf55891f2404b2c471932a37ea5b52842/adapters/bonbon.js#L38)
- [X] Move the repo to the `@npm` org
- [ ] Move the package to the `@npm` scope
- [ ] Make the package public, or give credentials to travis and our servers
- [ ] Add cachebusting support to acl-client ([in progress](https://github.com/npm/acl-client/blob/439bbd72ad9533f40eefc29e8dcf30fb05dbfb70/src/test/cache.js))

The client has lots of tests and a decent README. Please [have a look](https://www.npmjs.com/package/@zeke/acl-client). I expect to tackle cachebusting on the morrow.

zeke:
The code that's live on production right now fetches the logged-in user's profile from ACL (or redis once it's cached) for every single logged-in request. There are mocks for this in every handler test:

```js
userMock = nock("https://user-api-example.com")
    .get("/user/bob").times(3)
    .reply(200, users.bob)
```

This PR introduces a second request that fetches customer data from the license API. Rather than adding dozens more mocks to all those tests, I'm going to work on a mock-generating helper function that can be shared.

zeke:
Update: still slogging through all the broken tests. Halfway there.

zeke:
Update: most of the tests are passing in isolation, but the suite as whole still has lots of failures.

zeke:
I just tried running the tests with `mocha` instead of `lab`, and am still seeing lots of failures when running all tests together. This confirms that lab is not to blame: it's `nock`. I tried bumping `nock` from 1.4.0 to 1.6.0, but this didn't change anything.

My next move will be to sprinkle `nock.cleanAll()` in various places and see if it reduces error counts at the suite level.

zeke:
Lessons I've learned while working on this branch:

- Keep mock declarations (and state in general) out of `describe` blocks, as that seems to be the primary cause of tests that pass in isolation and fail as part of the full suite.
- Mocks that are used by multiple `it` tests should be declared in a `before` or `beforeEach` block.
- Always call the `done()` method on mocks to verify that they were used.
- Mocks should be local to the `it` test or `describe` block they apply to, so the test writer can reason about what network activity is taking place in that specific test.
- The `nock.cleanAll()` should generally not be used, as it allows incomplete mocks to be swept under the rug.
- nock's `.times(n)` should be avoided if possible, for the same reason.
- The www test suite runs in `mocha` with very little effort, but has the same nondeterministic execution order as `lab`.
- `mocha` seems to be way slower than `lab`. I'll produce a mocha PR with numbers when time permits.

PR #928, opened 2015-04-29
encourages password change after password reset, fixes #798

### LINKED ISSUE #798: Encourage password change after password reset
Here's the current flow on r2:

1. click the "forgot password" link
1. fill out username
1. click link in email
1. see a page with a password on it. Copy the password
1. click "log in" link in paragraph below
1. log in with username and copied password
1. get redirected to /~

This flow used to require (or at least encourage) an immediate password change, didn't it?

In addition to prompting for password change, I think this process could have fewer steps:

1. click the "forgot password" link
1. fill out username
1. click link in email
1. enter a new password
1. get redirected to /~

rockbot:
The key was to put the `mustChangePass` directive into the resource section, as it wasn't getting saved to user-acl.

Due to some user-acl limitations, we have to use a string instead of a usual boolean - fortunately we can nullify it, so `if user.resource.mustChangePass` still works just fine :-)

### DIFF
--- a/facets/user/show-forgot.js
+++ b/facets/user/show-forgot.js
@@ -60,7 +60,9 @@ function processToken(request, reply) {
         newAuth = {
           name: name,
           password: newPass,
-          mustChangePass: true
+          resource: {
+            mustChangePass: 'true' // must be a string, as per user-acl
+          }
         };
 
     request.logger.warn('About to change password', { name: name });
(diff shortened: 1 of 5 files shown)


zeke:
I just had a chance to try this out. The redirect bug is fixed, and we could :ship: it as-is.

I think the flow can be further improved, though. Couldn't the copying of the new password, clicking of the login link, and filling out of the login form all be skipped? Is there a reason we can't take the user directly to a "please enter and confirm your new password" screen, without those three steps?

PR #931, opened 2015-04-30
update dev instructions in readme

zeke:
cc @linclark, who is going through the process today.


linclark:
The way I finally resolved this was by blowing away my copy of newww and cloning it again. Is this the process we want to recommend?

zeke:
@linclark most users probably won't need to resort to that. I've added a blurb to the readme encouraging people to open issues if they have trouble.

PR #939, opened 2015-05-03
[WIP] CSP for Twitter conversion tracking

linclark:
In #913, the URL for Twitter was added to the CSP. However, the CSP is blocking execution.

csp

The error specifies multiple ways to fix this. I have done it the easiest way... however, we probably DO NOT want to do it this way. It would be much better to use nonces. There is an open issue to add nonce support to blankie, nlf/blankie#3.

### DIFF
--- a/lib/csp.js
+++ b/lib/csp.js
@@ -4,6 +4,7 @@ csp.default = {
   defaultSrc: 'self',
   scriptSrc: [
     'self',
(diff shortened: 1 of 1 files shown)


zeke:
Per an in-air discussion, we don't want to use `unsafe-inline`. Closing in favor of https://github.com/npm/newww/pull/946.

PR #941, opened 2015-05-04
Update nib link + grammar changes + trivial stylistic choices

MrNice:
Updated the link for the `nib` package to point to the current documentation.

Also removed some small grammatical errors, and for fun threw in some unnecessary stylistic changes.


MrNice:
Well that was the most painless PR

PR #946, opened 2015-05-04
annotate twitter codes

zeke:
so we have a local reference for the twitter pixel ids


zeke:
@linclark this is up on staging now and ready for review. Tests passing.

linclark:
Tested on staging. Can't tell yet whether the conversion came through, but I'll merge it and if it doesn't we can fix in another PR.

PR #953, opened 2015-05-05
Restore search url route #462

DanielBarnes:
with test.


aredridel:
What's the purpose of this change?

PR #955, opened 2015-05-05
contact: Send support requests directly to Zendesk

isaacs:
Review please: @rockbot 

In the future, we probably also want to send general inquiries and security tickets into Zendesk, but with different magic ticket flags so that we can route them appropriately.


isaacs:
This requires that we set a few fields in renv, as well.

Issue #948, opened 2015-05-04
stripe errors not all bubbled up to the user


ceejbot:
If, for instance, you enter your credit card expiration date incorrectly, Stripe will report an error & the payment will fail. However, the actionable information doesn't make it back up to the end user. They don't see that anything went wrong & are merely presented with the form again.

Let's make sure that errors are reported along with all relevant info so our beloved potential customer can give us some money.

PR #957, opened 2015-05-06
Bubble up stripe error

rockbot:
As per #948


seldo:
I don't think the error string should come directly out of Stripe. Stripe gives us specific error codes and we should handle them and use our own text for each one, plus a catch-all. Bubbling API error strings to the user can lead to very confusing/misleading error strings under edge conditions.

seldo:
This is the closest I can get to linking to the list of error codes: https://stripe.com/docs/api#errors

PR #964, opened 2015-05-08
allows starring of scoped packages (fixes #879)

### LINKED ISSUE #879: trouble starring scoped modules
### Observed Behavior

I am able to star scoped modules from the CLI, but when I attempt to star a scoped module on the website it does not seem to work.

### Expected Behavior

Starring scoped modules should work both in the CLI, and the website.

rockbot:
Not sure if this is the best solution, but essentially we were passing `@scope/package`, which was breaking the url we use to send the star to the user-acl. Instead, we should be sending `@scope%2fpackage` so that it doesn't seem like we're sending to a different url.

Using `encodeURIComponent(package)` would result in `%40scope%2Ffpackage`... would that work better?

### DIFF
--- a/models/package.js
+++ b/models/package.js
@@ -112,8 +112,9 @@ Package.prototype.count = function() {
 };
 
 Package.prototype.star = function (package) {
+
   var _this = this;
-  var url = fmt("%s/package/%s/star", _this.host, package);
+  var url = fmt("%s/package/%s/star", _this.host, encodeURIComponent(package));
   var opts = {
     url: url,
     json: true,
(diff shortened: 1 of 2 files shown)


zeke:
@rockbot the package presenter creates a `package.encodedName` property: https://github.com/npm/newww/blob/2bda66a400e0edf210e13e19f60cd47ad353a698/presenters/package.js#L15

models/package.js
```diff
@@ -112,6 +112,8 @@ Package.prototype.count = function() {
 };
 
 Package.prototype.star = function (package) {
+  package = encodeURIComponent(package);
```
ceejbot: This is a tweak, but I would encode the package name *only* in the url format line below, and use the unmangled name everywhere else. Redis and log lines don't care about slashes; only the uri you send to the user-acl cares.

PR #970, opened 2015-05-11
add coupon code handling

linclark:
This adds a coupon code field to the form and passes the coupon code to license api.


linclark:
@rockbot it looks like tests will not have completed before I leave for my talk. Feel free to merge when they finish.

PR #980, opened 2015-05-14
adds metrics to see hits/misses on readme caching

rockbot:


presenters/package.js
```diff
@@ -137,10 +138,25 @@ function processReadme (pkg) {
       }
 
       if (readme) {
+
+        metrics.metric({
+          name: 'readme.cache.hit',
```
ceejbot: Nitpicky, but there is method to my madness: call them `cache.readme.hit` and `cache.readme.miss`. This lets you put a whole bunch of cache-y metrics under one namespace. (I am assuming that this is more interesting than grouping a whole bunch of readme-y metrics.)

PR #983, opened 2015-05-20
cache is-paid status

rockbot:
It took a while due to yaks needing shaving, but finally we have cached isPaid information!

(The biggest issue was getting all the tests to pass - they should be doing so now.)

The included commits give a pretty good idea of what went down in the process of putting together this PR.

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -86,15 +86,7 @@ Cache.drop = function drop(opts, callback)
     assert(_.isObject(opts), 'you must pass a Request-ready options object to cache.drop()');
 
     var key = Cache._fingerprint(opts);
-    redis.del(key, function(err)
-    {
-        if (err)
-        {
-            Cache.logger.error('problem deleting ' + key + ' from redis @ ' + options.redis);
-            Cache.logger.error(err);
-        }
-        callback();
-    });
+    Cache.dropKey(key, callback);
 };
 
 Cache._getNoCache = function _getNoCache(opts, callback)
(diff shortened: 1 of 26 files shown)


lib/cache.js
```diff
@@ -63,10 +63,20 @@ Cache._fingerprint = function _fingerprint(obj)
...
 
+    return JSON.stringify(cleaned);
+};
+
+Cache._fingerprint = function _fingerprint(key)
+{
+    if (!_.isString(key)) {
+        key = this._cleanObj(key);
+    }
+
```
ceejbot: Nice touch here: hashes + prepends the name space but only mangles an object if needed.

lib/cache.js
```diff
@@ -147,10 +156,10 @@ Cache.setKey = function setKey(key, ttl, data, callback)
     }
 
     assert(_.isObject(redis), 'you must configure the redis client before using the cache.');
-    assert(_.isString(key), 'you must pass a string key');
+    // assert(_.isString(key), 'you must pass a string key');
```
ceejbot: Just delete the line if the assertion is no longer valid for the input.

lib/cache.js
```diff
@@ -199,7 +208,7 @@ Cache.dropKey = function dropKey(key, callback)
         return;
     }
 
-    redis.del(key, function(err)
+    redis.del(this._fingerprint(key), function(err)
```
ceejbot: Sweet.

seldo:
P.S. the work @bcoe is doing with @soldair to create a factory for mock objects in the license API tests might come in useful here.

PR #988, opened 2015-05-20
changes profile page messaging for paid users (fixes #965)

### LINKED ISSUE #965: "sign up for private modules" still shows up on profile page for users with private modules

rockbot:
with a test and everything :-)


seldo:
omg :sheep:

PR #989, opened 2015-05-20
disables private modules banner for logged-in, paid users (fixes #884)

### LINKED ISSUE #884: Disable private modules banner if current user is a paying customer
The preference is currently in localStorage, so a paying customer will see the banner on any new browser they use.

To avoid this, we'd need the user-fetching code to also hit the license API and cache the customer's payment status, then pass that to the DOM, preferably in a data attribute, much [like we currently do with the username](https://github.com/npm/newww/blob/021bdff48a22968f34612a9d147d097c5a8b64bc/templates/layouts/default.hbs#L76):

```handlebars
<a data-user-name="{{user.name}}" href="/~{{user.name}}"><img src="{{user.avatar.small}}" /></a>
```

rockbot:


seldo:
:sheep: :hamburger:

Issue #992, opened 2015-05-21
Link to multi license is broken


SimenB:
For example https://www.npmjs.com/package/JSONStream. Clicking the license 404s, as it tries to open http://opensource.org/licenses/(MIT%20OR%20Apache-2.0). What you'd open in this case I have no idea. Provide 2 links perhaps?

Also, the text is `(MIT OR Apache-2.0) license`, which isn't very human readable.

SimenB:
Pinging @kemitchell as requested in https://github.com/npm/normalize-license-data/issues/1 :smile:

kemitchell:
As much to remind myself as anything else:

- [ ] `spdx.parse(json)` for AST to translate to links and `<span>`s
- [ ] link to spdx.org/licenses/$identifier, not OSI

kemitchell:
Ongoing work in [spdx-to-html.js](https://github.com/kemitchell/spdx-to-html.js).

Krinkle:
Affecting https://www.npmjs.com/package/jshint. Currently renders a 404 Not Found to <http://opensource.org/licenses/(MIT%20AND%20JSON)>.

Change is https://github.com/jshint/jshint/commit/e47168f5aa.

aredridel:
We'll have to parse the expression with the spdx library and link more carefully. Thank you for the issue!

aredridel:
This is fixed by #996, thank you @kemitchell !

PR #996, opened 2015-05-22
Display SPDX license expressions and links

kemitchell:
This pull request harmonizes display of license links in the sidebars of package pages with recent clarifications to the `package.json` metadata guidelines. It directly addresses #992, and may or may not obviate npm/normalize-license-data#1.

Under the hood, the presenter for packages is changed to utilize [`spdx-to-html`](https://github.com/kemitchell/spdx-to-html.js), a purpose-built package, in lieu of `normalize-license-data`. `spdx-to-html` uses [`spdx`](https://github.com/kemitchell/spdx.js), the same package `normalize-package-data` and `init-package-json` now use for validation, to generate an AST for license expressions and flatten them into strings of HTML.

If there is anything icky about this patch, it's that `spdx-to-html` generates HTML that's inserted without escaping by Handlebars.

If there's work still to be done, it's for kemitchell/spdx-to-html.js#1, to make sure there's no wonky or misleading output if any when a package author decides to use SPDX expressions to their fullest. If expressions are new, have a look at [the README to `spdx.js`](http://npmjs.com/packages/spdx), which includes examples with multiple conjunctions and parentheses for grouping. There are a few multi-licensed packages in the registry that are doing basic conjunctions (many with my help). So far, I'm not aware of any using an expression that embodies this kind of edge case.

The arm of licensing law is a long arm indeed:

- npm/npm#8179 — here the guidelines were justly and ably amended
- [2.10.0 release notes](https://github.com/npm/npm/releases/tag/v2.10.0) — here that wise decision was announced to the people
- npm/normalize-license-data#1 — here the topic of spdx.org URLs was first broached

Poor @othiym23's been with me at each and every turn. @SimenB first saw the need for wiser license links, and acted on it. [JSONStream](http://npmjs.com/packages/JSONStream), I think it was.


kemitchell:
I've fiddled a bit with the code. Mind giving it one more shot?

I also rebased onto current master. Still no merge conflicts.

kemitchell:
Thoughts are these: I am not good at this!

SimenB:
Just a quick note: I tried to `require` the hbs-template and pass the data from @kemitchell's test through it, and the license is included in the html output, so it seems to work like it should.
Other than having to remove the partial `form_security` from the template to get it to run cleanly from a test, this is the full output from the test: http://hastebin.com/apokivameq.html. License is included on line 40.

To test it, I added `require('handlebars')` at the top of the test file, as it uses `require.extensions`, then `require(__dirname + "/../../templates/package/show.hbs")({'package': pkg})` in the test itself.

SimenB:
@rockbot Any news? :smile:

aredridel:
Looks like this branch could be rebased against current master -- I could take a peek at it but there's also conflicts.

kemitchell:
I will have another look this weekend, when I finish refactoring the SPDX-related modules. I also need to handle the magic values we recently coined for nonstandard licenses.

My main concern is that I'm really sure why my first patch didn't work. I'm still not entirely sure I understand how the template logic is actually rendered. I'll have another look.

aredridel:
Alright! Ping for help if you want a look -- happy to guide where I can.

kemitchell:
I've updated the PR and related modules to address the recent changes to valid license metadata. `npm test` is green with local Redis. I believe this is as far as I got before.

@rockbot, any chance you might run this on npm infrastructure?

sam-github:
Any chance of getting this fixed?

Its pretty broken at the moment, "SEE LICENSE IN LICENSE license" shows up on the web site, linking to http://opensource.org/licenses/SEE%20LICENSE%20IN%20LICENSE

/cc @rmg @superkhau

kemitchell:
Rebased. Green and clean.

rmg:
:+1: this would be especially nice since the npm cli currently gives warnings if you use the only method the website currently supports for fixing this URLs in package.

superkhau:
+1

presenters/package.js
```diff
@@ -22,9 +23,25 @@ module.exports = function(pkg) {
   pkg.scoped = pkg.name.charAt(0) === "@";
   pkg.encodedName = pkg.name.replace("/", "%2F");
 
-  pkg.license = normalizeLicenseData(pkg.license);
-  if (!pkg.license) {
+  if (pkg.versions && pkg.versions.indexOf(pkg.version) === -1) {
+    return Error('invalid pkg: ' + pkg.name);
+  }
+
+  if (!pkg.license || typeof pkg.license !== 'string') {
```
rockbot: Licenses can actually be an object, not just a string. For example, JSONStream has the following license:

```json
license: {
  "name": "(MIT OR Apache-2.0)",
  "url": "http://opensource.org/licenses/(MIT OR Apache-2.0)"
}
```

This line ends up deleting the license entirely in this case :-)

I'd rewrite this logic to consider either a single string or an object with the above properties.

kemitchell: @rockbot, the old Ruby-esque object style is now deprecated: https://github.com/npm/npm/blob/master/doc/files/package.json.md#license

Dominic was kind enough to update to the new format: https://github.com/npm/npm/blob/master/doc/files/package.json.md#license

cc @othiym23

SimenB: It should still be supported though, shouldn't it? npm has almost 200k packages (according to http://www.modulecounts.com/, no idea if it's actually accurate), I bet a huge chunk of them will not update to the new format, but they might use the old object or array-style.
Maybe add an annoying warning next to it, to encourage updating? But it shouldn't say `none license` (or just drop it entirely), or something similar when they actually define a license, albeit in a deprecated format.

kemitchell: There's a call to be made here.

The current behavior more strongly encourages conformance to the new guidelines.

@SimenB mentions a warning, which is (as usual!) great input. That's probably closest to the approach CLI has taken.

I'm cloning npm.dathub.com now, and will run a quick survey on Array and Object license props.

kemitchell: There is trouble in datland. I'll try to get it sorted. Stats on Array/Object license values may not be immediately forthcoming.

rockbot: this is totally an internal-to-npm failure on our part; the cli made that call (and rightly so), and other teams totally missed it. please consider the CLI docs as truth; we're going to fix the discrepancy on our end so that the data is normalized in our system along those same lines.

Unfortunately that will take some time, so I thank you for your continued patience as we reach congruency system-wide.

kemitchell: Thanks, @rockbot! Really appreciate your time on this.

kemitchell: Based on most recent npm.dathub.org, about 1,600 package have an Object for "license". Less than 200 have an Array.

Special thanks to @mafintosh for addressing maxogden/dat#408 so quickly.

kemitchell: Code for the survey: https://github.com/kemitchell/dat-npm-licenses

kemitchell:
Just checking in. Thanks again to all involved---y'all're busy wombats, I know!

If there's anything I can do independently to help this long, please let me know.

aredridel:
We may have some time coming up, probably not this week but just a bit after I hope :)

kemitchell:
Thanks, Aria!

kemitchell:
Just following up. Don't mean at all to nag.

Any chance full testing on non-npm folks' machines has come along since I PR'd? Thanks!

kemitchell:
The helper package I wrote for this PR had an errant `main` property in `package.json`. I've fixed it, pushed a patch release, and amended here to pick up the newer version. I'm not sure if that had anything to do with the failure here, but it's fixed now!

aredridel:
It does not fix it, but if you wanted to rebase against current master (`git fetch origin; git rebase origin/master` on this branch) I bet the tests would start passing, since master is in a good state right now and the failures seem to have come from the snapshot you developed this from.

kemitchell:
@aredridel I have been running the test suite with a local Redis. Is there any way for me to dev-deploy on my own machine to reproduce the problem you're seeing?

kemitchell:
I've rebased and run the test suite locally. There's a regression in my commit in unrelated tests. I'll look into it.

kemitchell:
Fixed the regression. Dropped some code from the package presenter that must have ridden along with all the rebases over time. The code was:

```javascript
// presenters/package.js
if (pkg.versions && pkg.versions.indexOf(pkg.version) === -1) {
  return Error('invalid pkg: ' + pkg.name);
}
```

kemitchell:
Green!

kemitchell:
Rebased. The only conflict was a package.json version bump.

PR #998, opened 2015-05-22
Add WOFF 2.0 version of the `icomoon` font

alrra:
The WOFF 2.0 file format brings some improvements in compression rates compared to WOFF 1.0.

See also:

  * [WOFF 2.0 Specification](http://www.w3.org/TR/WOFF2/)
  * [Browser support](http://caniuse.com/#feat=woff2)


alrra:
>  but I think the browser support isn't where I'd like to see it. 

@rockbot @ceejbot Browsers that do **not** have support for the WOFF 2.0 font file format will just grab the next file format they support (e.g.: `icomoon.woff` file). The only difference this change makes is that users with browsers that **do** support WOFF 2.0 will get a smaller font file (the `icomoon.woff2` file), everyone else will get the exact same as before.

(I do apologize for not being crystal clear the first time)

alrra:
Also note that: The media type with which `.woff2` font files are served does matter (see [discussion](https://github.com/h5bp/server-configs-apache/issues/32#issuecomment-46683006)), but since you're using `hapi.js` that in the end [uses](https://github.com/hapijs/mimos/blob/ec5ddbc02a44f287b6a48f3135ab3cacbe80d707/package.json#L17) [`mime-db`](https://github.com/jshttp/mime-db), you [should be covered](https://github.com/jshttp/mime-db/commit/a7cb52fe085bc3cf7666e0367da9db8af5cc59d6)!

PR #1004, opened 2015-05-28
Adds package privacy information

rockbot:
Know we know if a package is public...
screen shot 2015-05-27 at 4 57 03 pm

or private...
screen shot 2015-05-27 at 4 58 37 pm

And you can see it at a glance on your profile page, too!
screen shot 2015-05-27 at 4 59 17 pm


handlers/package.js
```diff
@@ -36,7 +37,7 @@ package.show = function(request, reply) {
         return;
       }
 
-      if (pkg.dependents.results.length) {
+      if (_.get(pkg, 'dependents.results.length')) {
```
ceejbot: nice way to avoid a giant if clause

models/user.js
```diff
@@ -198,6 +197,19 @@ User.prototype.getPackages = function(name, page, callback) {
         return reject(err);
       }
 
+      // it feels like this should really go in the handler instead,
+      // though we have client-side code
+      // (assets/scripts/fetch-packages.js) that needs this
+      // functionality as well... thoughts?
+      if (body.items) {
+        body.items = body.items.map(function (p) {
```
ceejbot: could just `each` them here, since you're not mutating the list, just a package field. Meh. No need to change.

PR #1007, opened 2015-05-28
Silence the cache logging in normal operations.

ceejbot:
Those info lines are now debug lines, so we don't spew a lot of logging on every cache operation. Also, we no longer uselessly log `[object Object]` as the cache key in those lines.

Updated the test that looked for the info log.

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -128,7 +128,7 @@ Cache.get = function get(opts, callback)
             return callback(null, value);
         }
 
-        Cache.logger.info('get: ' + opts.url);
+        Cache.logger.debug('get: ' + opts.url);
 
         Request(opts, function(err, response, data)
         {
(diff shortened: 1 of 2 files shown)


lib/cache.js
```diff
@@ -182,17 +182,19 @@ Cache.getKey = function getKey(key, callback)
...
         {
-            Cache.logger.error('problem getting ' + key + ' from redis @ ' + options.redis);
+            Cache.logger.error('problem getting ' + cacheKey + ' from redis @ ' + options.redis);
             Cache.logger.error(err);
         }
         else
         {
-            Cache.logger.info('retrieved ' + value + ' from ' + key);
+            Cache.logger.debug('retrieved ' + value + ' from ' + cacheKey);
         }
```
rockbot: :+1:

PR #747, opened 2015-03-18
Add Constructor.io autocomplete to search bar

danmcc:
Closes #720  Hi guys -- Co-founder of Constructor.io here. @EliFinkelshteyn and I have an autocomplete all ready for you. Currently, we have it auto-completing to packages for your users who know exactly what they're looking for, and to common searches for users who are looking more serendipitously.

Right now, we have you on the ad-based/branded plan, but we'd be happy to talk about taking you off of that and moving to something else, if you'd prefer.

Finally, we just made some basic styling for your current autocomplete, thinking you'd be likely to change it. That's very easy to do with css (in _header.styl), if you're interested. Otherwise, you're more than welcome to use everything as is. Please let us know if you have any questions, comments, or concerns!

PR #1010, opened 2015-05-29
Constructor.io

rockbot:
Follow up to #747 with slight bits of extra styling :-)

cc @EliFinkelshteyn @danmcc


danmcc:
Awesome!  Thanks, @rockbot!

PR #1013, opened 2015-05-30
Minor https related improvements

alrra:
1. Use `https://` instead of protocol-relative URLs
    
    Nowadays HTTPS is fast and increasingly necessary, so using protocol-relative URLs is becoming an anti-pattern, especially for CDNs.
   
    See also:
    
      * https://github.com/konklone/cdns-to-https#readme
      * https://developers.google.com/speed/libraries/devguide?csw=1#libraries:

2. Use `https://` where possible
    (I did check if the links do worked over https)


alrra:
Also, please change the URL from this project's description to use HTTPS.

alrra:
friendly ping @rockbot @jefflembeck

Issue #1020, opened 2015-06-02
Facebook preview shows a giant N


seldo:
Links to our content on Facebook look weird because our metadata tags are lacking:

![image](https://cloud.githubusercontent.com/assets/185893/7926277/09adea44-0885-11e5-9394-fd22d4270853.png)

We can do better, though this is obviously not a super high priority.

PR #1023, opened 2015-06-03
Improve Facebook preview (fix: #1020)

alrra:
* Change the `open-graph.png` image so that it has the recommended image size of `1200x630px`:

    > [Use images that are at least 1200 x 630 pixels for the best display on high resolution devices.](https://developers.facebook.com/docs/sharing/best-practices#images)

 * Update the `<meta property="og:image"...>` to include the full URL of the `open-graph.png` image as the [Open Graph Object Debugger](https://developers.facebook.com/tools/debug/og/object/) is complaining:

   > Object at URL 'https://www.npmjs.com/' of type 'website' is invalid because the given value '/static/images/touch-icons/open-graph.png' for property 'og:image:url' could not be parsed as type 'url' "

   

--

| Before | After | 
|:---:|:---:|
|  |  | 

--

Fix: npm/newww#1020


alrra:
> Do you need us to change the open-graph.png image? 

@rockbot No, I already did that (see: https://github.com/npm/newww/pull/1023/files). I just tend to write the [commit messages](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) in the imperative.

PR #1025, opened 2015-06-04
tweak Constructor.io background color

danmcc:
Hi guys,

We're glad to hear the autocomplete is working out great!  We're really excited to be providing the service.

We showed the branding edit you guys made to some people, and they didn’t notice our name, so we'd like to suggest one small tweak to make it slightly more prominent.  If we match the background color to the gray in your heading, we get something like this:

image

Is that cool?

### DIFF
--- a/assets/styles/_header.styl
+++ b/assets/styles/_header.styl
@@ -284,7 +284,7 @@ nav
   border-color greigh5
 
 #powered-by-constructor-io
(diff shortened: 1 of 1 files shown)


danmcc:
Ah good call @rockbot !  I hadn't noticed that.  All set.  Thanks!

PR #1027, opened 2015-06-06
Update assets/scripts/keyboard-shortcuts.js

escaton:
Add "f" shortcut to acces search.


escaton:
@rockbot just to be similar to system `( cmd | ctrl )+f`

PR #1042, opened 2015-06-16
feature parity with the updated license api

rockbot:
The license API has changed a bit; this PR allows us to launch the license API to production without affecting users while we work on the next bits.

cc @bcoe

### DIFF
--- a/handlers/customer.js
+++ b/handlers/customer.js
@@ -1,14 +1,13 @@
-var billing = module.exports = {};
+var customer = module.exports = {};
 var utils = require('../lib/utils');
 
-billing.getBillingInfo = function (request, reply) {
+customer.getBillingInfo = function (request, reply) {
 
   if (!request.features.billing_page) {
     return reply.view('errors/not-found').code(404);
   }
 
   var opts = {
-    namespace: 'billing',
     title: 'Billing',
     updated: ('updated' in request.query),
     canceled: ('canceled' in request.query),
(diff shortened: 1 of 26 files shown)


jefflembeck:
:sheep: it.

handlers/customer.js
```diff
@@ -91,3 +102,25 @@ billing.deleteBillingInfo = function(request, reply) {
     return reply.redirect('/settings/billing?canceled=1');
   });
 };
+
+// var plans = {
```
bcoe: can we delete this commented code :+1:

handlers/customer.js
```diff
@@ -64,25 +65,35 @@ billing.updateBillingInfo = function(request, reply) {
       email: billingInfo.email
     };
 
-    sendToHubspot(process.env.HUBSPOT_FORM_PRIVATE_NPM_SIGNUP, data, function (er) {
-      if (er) {
-        request.logger.error('unable to send billing email to HubSpot');
-        request.logger.error(er);
+    var planInfo = { plan: 'npm-paid-individual-user-7' };
+
+    request.customer.updateSubscription(planInfo, function (err, unused) {
```
bcoe: It might be good to call this `createSubscription`, since it's always attempting to create a subscription rather than updating an existing one :+1:

bcoe:
This is looking good. If a user updates their billing information we currently always try to create a subscription for npm-private-modules, and we just allow an error to occur and log it? you'll get an error because the customer already has a subscription -- we should test this flow.

Issue #991, opened 2015-05-21
Link color inside a header


simov:
:wave:

If you have a markdown like this

```md
## 100+ Supported Providers / [OAuth Playground][playground]
```

It will result in a link with a blue color inside the header when rendered in GitHub https://github.com/simov/grant

But in NPM there is no way for the user to know that this was actually a link https://www.npmjs.com/package/grant so it's very easy to miss that information (and that's quite important in my case :)

Gioyik:
I created a PR (#1049) proposal for this issue. Let me know if is good.

aredridel:
Awaiting response on #1049; we'd certainly love that approach.

karanjthakkar:
@aredridel Can this be closed since proposal #1049 has been closed? Or is this likely to be an issue even after the changes mentioned in https://github.com/npm/newww/pull/1049#issuecomment-183523877 are done?

PR #1049, opened 2015-06-19
Link color inside a header

Gioyik:
This is PR proposal for #991. Let me know if is OK or needs a change. :)


simov:
Thanks for your work on this one @Gioyik :+1: 

I'm not sure if this is going to make all of the headers red or just the headers with additional link in them.

Probably additional selector will be required, but I can't tell just by looking at the changed line.

Gioyik:
@rockbot yes, you're totally right, most of the `## (h2)` are rendered as links. I found a solution using Javascript that could work for everybody. Check this [jsfiddle](http://jsfiddle.net/Gioyik/vzahskny/) as test case and let me know if you think it is OK as solution for this PR.

Thanks a lot

simov:
Thanks for the example, @Gioyik. I think a pure css solution would be better in this case, check out this [fiddle](http://jsfiddle.net/simo/6g3k99jq/).

Either

```css
h2:not(.deep-link) a { color: red; }
```

or

```css
h2 a:not([href^="#"]) { color: red; }
```

works. No idea which one is more optimal, but the second one seems more reliable (in case of html change).

Gioyik:
@rockbot yes, both will work without breaking other package or section in the site. Just let me know which you like more or think is better (second css solution from @simov or mine with Javascript) and I will update the PR with it.

Thank you @simov. I didn't know about that possibility using CSS, I was looking for a css solution first but didn't got it.

simov:
@rockbot well that's just CSS.

My assumption was that each header have an anchor link in it with a `.deep-link` class set on it, or even the more obvious - that link's `href` starts with an `#` character since it's an anchor.

So the above selectors match all links that are inside `h2` tag, excluding the anchor link.

Of course this is going to work only for `h2` tags, if we want it to work for all other header types, we should state that explicitly:

```css
h1 a:not([href^="#"]) { color: red; }
h2 a:not([href^="#"]) { color: red; }
h3 a:not([href^="#"]) { color: red; }
```

And that's not specific to a project, but rather specific to HTML structure, or like in the above selector, specific to attribute value.

If for some reason your markdown parser starts to generate the anchor's `href` attributes like `href="/package/grant#header"` all links will become red. But that's highly unlikely IMO.

aredridel:
I do like the latest proposed solution; if you want to update your PR we can get that merged.

pavelloz:
Maybe you will like also my oldschool view on the subject -- why not just add the `underline` for the `a:link` inside headings?

It will be very simple to implement, it is widely understood what underlined piece of text means and it doesn't require introduction of new color in headings.

stubbornella:
I'm going to close this because we aren't going to change our header colors right now. We have some design changes in the pipe that it would likely conflict with.

PR #1050, opened 2015-06-19
Issue 1043

### LINKED ISSUE #1043: /private-modules should be aware of paid users
The signup form should go away if you are a paid private modules user; the tutorial and instructions should stay.

jefflembeck:
Quick script addition to remove part of static page.

Fixes #1043

### DIFF
--- a/assets/scripts/remove-private-module-signup.js
+++ b/assets/scripts/remove-private-module-signup.js
@@ -1,0 +1,8 @@
+module.exports = function(){
+  $(function () {
+    var loggedIn = $('#user-info a').has('[data-is-paid=true]');
+    if( loggedIn ){
+      $('.private-module-info-container').remove();
+    }
+  });
+};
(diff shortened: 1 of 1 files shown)


assets/scripts/remove-private-module-signup.js
```diff
@@ -0,0 +1,8 @@
+module.exports = function(){
+  $(function () {
+    var privateModuleCustomer = $('#user-info a').is('[data-is-paid=true]');
```
rockbot: wouldn't `$('#user-info a').data('is-paid')` also work? I think that's what we're using elsewhere in the codebase, but if you've got Reasons™ for this format, I'm open to learning :-D

jefflembeck: Yup. That'd work too.

I tend to stay away from using `$.fn.data` because of old speed reasons (it's slow, but in all reality, not much slower than `.is`), but in this case, it wouldn't matter at all. I'll make the change.

That being said, if we're ever in a tight loop and need to use `.data`, we should use `$.data` over `$.fn.data` as it's significantly faster and has the exact same results :)

That being said, I should totally make that `$.data( '#user-info a', 'is-paid' )'`

PR #1051, opened 2015-06-19
Overzealous regex was logging people out

### LINKED ISSUE #1032: User authentication becomes invalided when viewing certain package pages.
Behavior: 

- A user logs in to their account.
- A logged in user visits a package page with `static` in the name.
- The users authentication becomes invalidated when on that particular package. 

Perhaps the word `static` in a package name is causing us to cache resources that should not be cached.

jefflembeck:
If a package had 'static' in the name, the url wasn't being
authenticated against. We'd like to not go through the process of
AUTHing requests for static files, but of course don't want to log
people out because they look at a package that has 'static' as part of
the title.

Fixes #1032

### DIFF
--- a/adapters/bonbon.js
+++ b/adapters/bonbon.js
@@ -35,7 +35,7 @@ exports.register = function(server, options, next) {
       start: Date.now(),
     };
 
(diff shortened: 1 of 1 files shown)


jefflembeck:
yes!

PR #1056, opened 2015-06-23
Update README.md

ckross01:
adjusted the documentation for where the routes are located to the folder they now exist in. Reference a .js file that is no longer there (Routes.js).


jefflembeck:
Thanks!

PR #1058, opened 2015-06-23
Add npm-tips to login page.

### LINKED ISSUE #1047: Replace "You look lovely" with `npm-tips` module
Continuation from https://github.com/npm/newww/issues/1046

https://github.com/npm/npm-tips

Use the `npm-tips` module for the login screen, and anywhere else that we feel like throwing a little bit of random surprise helpfulness.

jefflembeck:
Fix #1047

- Add tips
- Add marky-markdown support
- Compliment removed

### DIFF
--- a/facets/user/show-login.js
+++ b/facets/user/show-login.js
@@ -118,7 +118,7 @@ module.exports = function login (request, reply) {
   if (request.method === 'get' || opts.error) {
     request.timing.page = 'login';
     request.metrics.metric({name: 'login'});
-    opts.subtitle = marky(tips()).html();
+    opts.tip = marky(tips()).html();
     return reply.view('user/login', opts).code(opts.error ? 400 : 200);
   }
 };
--- a/facets/user/show-login.js
+++ b/facets/user/show-login.js
@@ -3,7 +3,9 @@ var Boom = require('boom'),
     fmt = require("util").format,
     redis = require("../../adapters/redis-sessions"),
     avatar = require("../../lib/avatar"),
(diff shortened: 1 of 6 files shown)


facets/user/show-login.js
```diff
@@ -116,6 +118,7 @@ module.exports = function login (request, reply) {
   if (request.method === 'get' || opts.error) {
     request.timing.page = 'login';
     request.metrics.metric({name: 'login'});
+    opts.subtitle = marky(tips()).html();
```
rockbot: do we need to pass a parameter to `tips()`, as per https://github.com/npm/npm-tips/blob/master/index.js#L9 ?

locales/en_US/user.js
```diff
@@ -17,7 +17,6 @@ module.exports = {
 
   login: {
     title: "login",
-    subtitle: "you look lovely today, btw.",
```
rockbot: food for thought: npm-tips may become a bit of an issue once we bring in internationalization. not something we need to worry about right now, but definitely something to keep in the backs of our minds.

facets/user/show-login.js
jefflembeck: If we do, it doesn't come back random.

locales/en_US/user.js
isaacs: Maybe when we start internationalizing the site, we can also i18n the npm-tips module?

jefflembeck: replaced with "ohai" for now - and that's used as the subtitle

PR #1061, opened 2015-06-24
Add titles for browse pages

BrianCS:
I opened a few tabs and couldn't keep them straight. Here are proposed page titles for "/browse/keyword/{keyword}", "/browse/depended", "/browse/depended/{package}", "/browse/star", "/browse/updated", and "/browse/created" pages.

This also removes a search results page-specific way of displaying search queries in template logic in favor of building the title earlier.

Since #761 is still open this was done blind 🙀, but think I got 'i's dotted and 't's crossed

### DIFF
--- a/facets/registry/show-search.js
+++ b/facets/registry/show-search.js
@@ -63,7 +63,7 @@ module.exports = function (request, reply) {
     request.metrics.metric({ name: 'search', search: request.query.q });
 
     merge(opts, {
-      title: 'results for ',
+      title: 'results for '+request.query.q,
       page: page,
       q: request.query.q,
       results: response.hits.hits,
--- a/handlers/browse.js
+++ b/handlers/browse.js
@@ -14,6 +14,7 @@ browse.packagesByKeyword = function(request, reply) {
   }
   var options = {
     keyword: context.keyword,
+    title: 'packages with keyword \'' + context.keyword + '\'',
     count: defaultCount,
     offset: Math.abs(parseInt(request.query.offset, 10)) || 0
   }
@@ -31,6 +32,7 @@ browse.mostDependedUponPackages = function(request, reply) {
   var context = {}
   var options = {
     sort: 'dependents',
+    title: 'most depended upon packages',
     count: defaultCount,
     offset: Math.abs(parseInt(request.query.offset)) || 0
   }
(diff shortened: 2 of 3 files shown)


BrianCS:
Hm, two possible approaches:

1. Maybe just give Handlers a i18n function to call?
 * The convention with Hapi's flavor of Handlebars layouts seems to be that `<title>` is coming from Controllers (I'm a total Hapi beginner so I'm crudely guessing by browsing over various blog posts), and the convention with Handlebars seems to be "when in doubt maybe avoid any complexity in templates and increase in controllers-ish parts" (suggested by things like "Handlebars doesn't have an else-if"). So maybe this bit of text coming from upstream is okay.
   * But a single source really probably would be nice.
 * Other than its folder path `t.js` isn't anything Handlebars Helper specific, so it could move to some other directory, then some facade could be added to the helpers folder so Hapi autowires it (`module.exports = require('../../new/path/to/t.js')`?)
 * …then Handlers could `require` the new file's now-shared translation utility and start e.g. setting {{title}} by riffling through the i18n objects.

2. Have Handlers assign some new generic identifier, `i18nRoot` or `templateId` or something, extend `t.js` to first go look for a translation key nested under that path first, then start nesting `title`s there and letting the default Layout find them contextually.
  * It could go something like
      1. e.g. in `browse.js`'s `browse.mostDependedUponPackages` provide a context value like `i18nRoot: 'browse.depended'` and render the view.
      2. Then write out title like `<title>{{t title}}</title>`
      3. Then have the `t` helper grab the context identifier ("browse.depended"), the string identifier ("title"), combine them and go and look for the title at `locales.en_US.browse.depended.title`
        * and if that doesn't exist then go try to match by string identifier alone (so shared strings can exist for bits used across multiple contexts)
  * That keeps user-facing strings strictly in those i18n locales objects, but does add an idiosyncratic new quasi-required context property to set for basically every view.
    * but on the _extremely minor_ bright side `{{t}}` calls will on average take a couple fewer keystrokes

No. 2 smells workable to me; also I have it basically working in a strawman sort of way in a scratch project.

I haven't thought about this bit much yet but the translation helper might need to be able to execute Handlebars logic found in locales strings (seems like it doesn't now but honestly haven't checked), since eventually e.g. package names will need to appear in titles and different languages will put them in different parts of a phrase. If I really want these titles I think that has to be considered too.

Is there an open issue for the "how to translate titles" thing? I can go catch up there

PR #1070, opened 2015-06-26
Issue 1069

jefflembeck:
Maybe we keep this open for a few, while I continue to audit.


aredridel:
I just rebased this branch onto current `master`, so commit f670c51 replaces commit 4929203

PR #1071, opened 2015-06-26
Add sudo permissions

nhubbard:
For pull request 1070, the build complained that it could not get sudo permissions but still passed. I aim to fix that with this.


ngoldman:
`sudo: false` enables container-based infrastructure on travis.

http://docs.travis-ci.com/user/workers/container-based-infrastructure/

I think this may be why it was set to false at 415cc055e5e04b862bf0a80b7fbc3bf86430a1b1. The complaint you mention is just a warning.

:grin:

aredridel:
I'm going to close this, since the build works as is, and on the faster, container-based infrastructure, and without using a git master version of the npm cli.

PR #1026, opened 2015-06-05
Make search prettier

rockbot:
In this PR you will find a wholly redone layout of search:
* Gone are the triple columns, back to a more readable single-column (with a sidebar), as per https://github.com/npm/newww/issues/448. Yes, there's still a lot of whitespace. We've got ideas on how to fix it, but we're going to iterate on it.
* Also gone are the download counts. They're not particularly useful right now, since the numbers in ElasticSearch don't actually match the numbers in the downloads api. So let's get rid of the confusion. (re: #394)
* The star counts are back, though. Hopefully this will give users some notion of popularity that the downloads count may have provided before. (Again, we can totally iterate on this.)
* Keywords are back, too. They're not clickable links yet, but they're pretty helpful in determining if the package is more like what you're looking for.
* And know you know who the author is, too. (As per #239)

In addtion, I had to remove the .woff2 font file, since every conversion of the updated .woff/.ttf/.eot rendered a broken .woff2. I'm sad about the performance loss, but broken fonts are not okay. (cc: @alrra)

PR #1073, opened 2015-06-29
Readd woff2 src descriptor to the @font-face rule

alrra:
The woff2 `src` descriptor was removed from the @font-face rule for the reason that the woff2 font was broken in Firefox Developer Edition, however, I personally couldn't reproduced that issue (see [test page](https://dl.dropboxusercontent.com/u/479990/github/tests/npm/newww/1026/index.html)).

--

Previous discussion:  https://github.com/npm/newww/pull/1026#issuecomment-109395406


jefflembeck:
I'm into it, and tbh, there is a good feature test for woff2. Let's put the feature test in first (as to support as many browsers as possible) and then look at adding back in the woff2 (huge decrease in initial font loading)

alrra:
> there is a good feature test for woff2. Let's put the feature test in first (as to support as many browsers as possible)

@jefflembeck I do apologize, but I do not understand what exactly do you mean by that? If you mean a feature detection test, you don't need such a thing (see [comment](https://github.com/npm/newww/pull/998#issuecomment-106569633)).

jefflembeck:
oh, @alrra, sorry - brain hiccup. You're right - I've been planning on loading our fonts async, so as to avoid FOIT -- [see many of @zachleat's works on the subject referenced here](https://css-tricks.com/fout-foit-foft/).

I can likely address this with something like [FontFaceObserver](https://github.com/bramstein/fontfaceobserver)

alrra:
> oh, @alrra, sorry - brain hiccup. You're right -

@jefflembeck Don't worry! :purple_heart: 

--

> I've been planning on loading our fonts async, so as to avoid FOIT -- see many of @zachleat's works on the subject referenced here.
>
> I can likely address this with something like FontFaceObserver

@jefflembeck I don't know what the browser support is for `npmjs.com`, but from what I can tell, you [mainly support browsers that have support for SVG](https://cloud.githubusercontent.com/assets/1223565/8416781/ab8fe754-1eb0-11e5-80da-72f60580c647.png)). So, I've been wondering: why not just [go with SVG icons instead of icon fonts](https://css-tricks.com/icon-fonts-vs-svg/)?

alrra:
@jefflembeck see also: [_Death to Icon Fonts_](https://speakerdeck.com/ninjanails/death-to-icon-fonts)

jefflembeck:
@alrra I actually plan on addressing that we use icon fonts at  some point. I might be really biased, but https://github.com/filamentgroup/grunticon-lib is a pretty dang good solution for this problem.

We'll see. It's possible to use icon fonts in a pretty dang robust manner as well.

alrra:
> We'll see.  

@jefflembeck Sure. :)

>  I actually plan on addressing that we use icon fonts at some point.

Coming back to the change proposed by this pull request...  when you have time, can you look over it (namely test that the WOFF2 font file also works for you), and decide if you want to merge it? Thanks!

jefflembeck:
@alrra going to be testing this like crazy today

jefflembeck:
woo https://github.com/npm/newww/commit/fb16007d0224f4e5e1a84194dbdec939ef9d383d - it's in

alrra:
> woo fb16007 - it's in

:tada:

PR #1090, opened 2015-07-14
NOW WE CAN REMOTE LIKE REAL PEOPLE OMG

rockbot:
@jefflembeck @ceejbot please scrutinize this PR. I don't think it we will have any production problems if we merge this code in, but I'd rather be safe than sorry.


jefflembeck:
I'm ok with those things!

Issue #1066, opened 2015-06-25
Exact match searches don't show up first in the results


polson136:
When searching for a specific package, say "sequelize", the "sequelize" package does not show up in the first page of results. Granted, you could select it from the drop down box, but if you are fast at typing, it is possible to enter your query before the drop down appears, therefore it should not be a substitute for the search results. The exact match should be either at the top of the result list or represented in some other way.

flyswatter:
Yeah, I found the same is true for many other search queries, like "tape", where I searched the first 12 pages of results to not find the module whose name was an exact match!

zhanzhenzhen:
When typing "bluebird" the famous bluebird package is even at the second page. This is weird...

rreusser:
Are +1's useful or just irritating?

Only wanted to register that this really seems to affect discoverability. As a stopgap, even just sort options (downloads, dependent count, stars, ?) would be extremely helpful.

'bluebird' is the first result in the dropdown if you type 'bluebird', but as a random counterexample to the dropdown's usefulness, 'euclideanmst' doesn't return a module with the exact name.

rreusser:
Looks great! Much appreciated.

zhanzhenzhen:
Looks like name with a dash such as "coffee-script" still has this problem.

jefflembeck:
@zhanzhenzhen Thanks! This is a known issue and there is a bug filed for it - https://github.com/npm/newww/issues/1103. We're iterating and making the search better piece by piece.

zhanzhenzhen:
This bug has been fixed. But now I've found an exception:
Type "proof".
The exact match is only at the second position.

focusaurus:
I have found another exception today, `joi`. Has this regressed?

flyswatter:
I also think this has regressed in some way. For example, [bluebird](https://www.npmjs.com/search?q=bluebird) is again not returned first for an exact query.

evanjbowling:
I just noticed this as recent as April 16. I was searching for "socket.io" and it came up with different results.

Reproduce by opening: https://www.npmjs.com/search?q=socket.io
Observation: It returns either "socket.io" or "socket.io-zeromq" first.

Possible related issues I found: #1103, #1096, #1630 
While this isn't ideal, it didn't really prevent me from finding the right package.

flyswatter:
@rockbot this should be reopened.

PR #1096, opened 2015-07-17
packages with the searched name are definitely at the top now

### LINKED ISSUE #1029: exact match not listed on first page of search results
A keyword search such as https://www.npmjs.com/search?q=stylus does not list the exact match "stylus" package on the first page of results. It's on page 5. I'm pretty sure this is a regression as I recall this working in the past.

rockbot:
fixes #1029 and #1066 

when we updated search we forgot to push the exact-name matches to the top. this is now fixed :-)

### DIFF
--- a/facets/registry/show-search.js
+++ b/facets/registry/show-search.js
@@ -23,7 +23,7 @@ module.exports = function (request, reply) {
       size : perPage,
       "query" : {
         "dis_max": {
-          "tie_breaker": 0.7,
+          "tie_breaker": 0.9,
           "boost": 1.2,
           "queries": [
             {
@@ -52,7 +52,7 @@ module.exports = function (request, reply) {
                   {"match_phrase": {"readme": request.query.q} }
                 ],
                 "minimum_should_match": 1,
-                "boost": 50
+                "boost": 30
               }
             },
           ]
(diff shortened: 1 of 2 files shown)


feross:
NICE!

jden:
:+1:

noopkat:
🎉❤️

feross:
Another improvement would be to demote deprecated packages.
https://staging.npmjs.com/search?q=webtorrent shows "webtorrent-tracker"
and "webtorrent-swarm" which are both deprecated now.

facets/registry/show-search.js
```diff
@@ -22,22 +22,41 @@ module.exports = function (request, reply) {
...
-                    boost: 20
+        "dis_max": {
+          "tie_breaker": 0.9,
+          "boost": 1.2,
+          "queries": [
+            {
+              "function_score": {
+                "query": {
+                  "match": {
+                    "name.untouched": request.query.q
```
ceejbot: Boommity boom. The thing I like: that this is the only tinkering with the result scores.

jefflembeck:
results for grunt-grunticon 2015-07-20 08-37-06

So, interesting thing, searching for something that has packages with same term, but with a hyphenated suffix, seems to yield interesting results

jefflembeck:
:sheep:

jefflembeck:
yay

PR #1099, opened 2015-07-20
Build file tota11y

jefflembeck:
I know this is small but it's a separation of where this process belongs (not in staging/prod builds).


jefflembeck:
Oof. Built off of #1098

PR #1074, opened 2015-06-29
Make clear button be accessible on small screens

alrra:
On small screens the search input slides under the submit button, thus, preventing the clear button (provided by some browsers) to be accessible.

| Before | After |
|:---:|:---:|
|  |  |

| Before | After |
|:---:|:---:|
|  |  |

PR #1101, opened 2015-07-20
Digest

jefflembeck:
We were having problems with cache on Fastly.

Let's engage in some md5-based cache busting!


alrra:
> We were having problems with cache on Fastly.

Probably related: https://github.com/npm/newww/pull/1074#issuecomment-120141716.

jefflembeck:
(not working with `watch` yet, fixing)

jefflembeck:
Gonna find another solution.

PR #1109, opened 2015-07-23
Link to policies page so we can add N additional policies

seldo:
1. We added the recruitment policy
2. Adding an extra link messes up the symmetry of the footer
3. We'll probably want to add extra policies
4. It's okay if the trademark policy is 1 click further away

Conclusion: this PR.


jefflembeck:
:sheep:

PR #1112, opened 2015-07-29
Fixed issue #43: www sessions are not correctly expired in redis on logout

ceejbot:
Hang the session id field onto the `loggedInUser` object so the value is available later on when attempting to act on the session in redis.

No tests yet, because the login/logout code has no existing tests to build on, and shaving that yak is a task for another day.

### DIFF
--- a/adapters/bonbon.js
+++ b/adapters/bonbon.js
@@ -39,6 +39,8 @@ exports.register = function(server, options, next) {
     if (request.auth && request.auth.credentials && !request.path.match(/static\//)) {
       UserModel.new(request).get(request.auth.credentials.name, function(err, user) {
         if (err) { request.logger.warn(err); }
+        user = user || {};
+        user.sid = request.auth.credentials.sid;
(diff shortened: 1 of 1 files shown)


adapters/bonbon.js
```diff
@@ -39,6 +39,8 @@ exports.register = function(server, options, next) {
     if (request.auth && request.auth.credentials && !request.path.match(/static\//)) {
       UserModel.new(request).get(request.auth.credentials.name, function(err, user) {
         if (err) { request.logger.warn(err); }
+        user = user || {};
+        user.sid = request.auth.credentials.sid;
         request.loggedInUser = user;
         request.customer = user && new CustomerModel(user.name);
```
rockbot: if `user` ends up being an empty object in line 42, then `user.name` will break. (which may also be part of the reason why things are broken now).

could we rewrite this line to `request.customer = user.name && new CustomerModel(user.name)` (since `user` _will_ be defined by this point)?

PR #1115, opened 2015-07-30
Improve Stylus' build instructions

diessica:
There isn't any "static" folder in this repo (cause it's being ignored in .gitignore file), so a link to this folder is actually a broken link.


zeke:
:+1:

PR #1116, opened 2015-07-31
Fix #145

jefflembeck:
In this, there were some pretty heinous CSS crimes committed.

I'm sorry for these. We will come back to this technical debt during the
redesign

This should fix #145

### DIFF
--- a/templates/layouts/default.hbs
+++ b/templates/layouts/default.hbs
@@ -75,7 +75,7 @@
           </a>
           <ul>
             <li class="username"><a href="/~{{user.name}}">{{user.name}}</a></li>
-            <li class="logout"><a href="/logout" class="subdued">log out</a></li>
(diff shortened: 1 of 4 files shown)


templates/layouts/default.hbs
```diff
@@ -75,7 +75,7 @@
           </a>
           <ul>
             <li class="username"><a href="/~{{user.name}}">{{user.name}}</a></li>
-            <li class="logout"><a href="/logout" class="subdued">log out</a></li>
+            <li class="logout"><form method="POST" action="/logout">{{> form_security }}<button type="submit" class="subdued"/>log out</button></form></li>
```
rockbot: I don't think you want that extra `/` in `<button ... />`, do you?

jefflembeck: Good call

Issue #812, opened 2015-04-03
display count beside dependents on package page


zeke:
Just like dependencies

PR #1126, opened 2015-08-05
Re-add dependents count to package sidebar

mikaelbr:
It seems as the fix from #812 has been lost in some refactoring of the templates. This fix re-adds the dependents count to the sidebar.

Couldn't verify the fix as I'm unable to run the dev server locally (#761).

### DIFF
--- a/templates/package/show.hbs
+++ b/templates/package/show.hbs
@@ -186,7 +186,7 @@
   </p>
 
   {{#if dependents.results}}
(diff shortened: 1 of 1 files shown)


zeke:
> Couldn't verify the fix as I'm unable to run the dev server locally

:cry:

aredridel:
This actually can't work (shows "NaN more") because the backend doesn't know how many there are -- closure over a search means doing all the potentially large work of a search, then discarding most of the result, and our backend optimizes and just returns a scattering.

PR #1129, opened 2015-08-07
Add already paid users

rockbot:
users can only be paid for once, whether that's by themselves or by one of the orgs of which they're a member. this PR lets us catch when a user has already been paid for, but doesn't stop them from being added to an org.


jefflembeck:
:sheep:

PR #1131, opened 2015-08-07
Org management page

jefflembeck:


assets/scripts/index.js
```diff
@@ -1,4 +1,4 @@
-window.$ = require("jquery");
+window.$ = window.jQuery = require("jquery");
```
rockbot: is this for the 3rd party JS?

jefflembeck: Yes. Though I suggested a fix in the script that necessitates us doing that, and they said they'll get right on it.

jefflembeck:
Current status: fixing all of the broken tests.

assets/scripts/index.js
rockbot: :+1:

agents/org.js
```diff
@@ -71,7 +74,12 @@ Org.prototype.get = function (name, callback) {
     var org = {};
 
     org.info = results.org;
-    org.users = results.users.items;
+    org.users = results.users;
+    org.users.items = results.users.items.map(function(user) {
+                  user.avatar = avatar(user.email);
+                  return user;
+                });
+    org.packages = results.packages;
```
rockbot: are we planning to display packages? or is this for some time in the future? (mostly just curious right now)

jefflembeck: yes to both. We're going to show them as read-only if time allows to add it. If not, we'll be adding them in an editable fashion in the future.

rockbot: :+1:

assets/styles/org.styl
```diff
@@ -0,0 +1,56 @@
+.billing-header
+  text-align left
+
+.org-table
+  border 1px solid #ababab
```
rockbot: do we want to assign `#ababab` to a variable name? (see variables.styl) it makes me nervous to see random color hashes :-)

handlers/org.js
```diff
@@ -6,29 +6,46 @@ exports.getOrg = function (request, reply) {
...
+        return reply.view('errors/not-found', err);
+      } else {
+        return reply.view('errors/internal', err);
+      }
+    }
     opts.org = org;
 
     Customer(loggedInUser)
-      .getSubscriptions(function (err, subscriptions) {
+      .fetch(request.loggedInUser.email, function (err, cust) {
```
rockbot: it seems a bit weird to send the `loggedInUser` to the Customer model and then also send the `fetch` function `request.loggedInUser.email`, no?

lib/csp.js
```diff
@@ -58,7 +61,9 @@ csp.default = {
   frameSrc: [
     'https://checkout.stripe.com',
     'https://js.stripe.com',
-    'https://www.youtube.com'
+    'https://www.youtube.com',
+    'https://s3-us-west-2.amazonaws.com/files.accountdock.com/pre.prod.html',
+    'https://accountdock.com/app',
```
rockbot: +1 to trailing commas :-)

models/customer.js
```diff
@@ -16,6 +16,33 @@ var Customer = module.exports = function(name, opts) {
   }, opts);
 };
 
+Customer.prototype.fetch = function(id, callback) {
```
rockbot: `fetch` vs `get` is a bit confusing, naming-wise. i understand that they're getting different things, but I wonder if there's a better way to name these two functions for clarity?

templates/org/info.hbs
```diff
@@ -1,53 +1,160 @@
...
-          <input type="submit" class="full-width" value="Add User" />
-
-        </form>
+  <div class="tabs">
+    <div class="tab-nav-container">
+      <div class="nav-3-container">
+        <ul class="nav-3 clear" role="navigation">
+          <li class="current"><a href="#org-info-tab-1">Overview</a></li>
+          <li><a href="#org-info-tab-2">{{org.users.count}} Users</a></li>
+          <li><button type="button" id="billing" data-stripe-id="{{org.customer_id}}">Payment Information</button></li>
```
rockbot: why a button?

templates/org/info.hbs
```diff
@@ -1,53 +1,160 @@
...
+              </td>
+            </tr>
+          {{/each}}
+          <tr>
+            <td colspan=2>
+              <form method="POST" class="org-user-add" action="/org/{{org.info.name}}">
+                <label for="name"><span class="a11y-only">Username</span></label>
+                <input required="required" type="text" id="username" name="username" placeholder="Invite a Member"/>
+                <fieldset>
+                  <legend><span class="a11y-only">User Type:</span></legend>
```
rockbot: i love this use of the legend. THANK YOU for leveling up our a11y!!

templates/org/info.hbs
jefflembeck: It's a tiny thing, but I've had better experience with browser-related bugs and styling `button`s over `input type=submit`. Also, I have a bit of a habit from sometimes needing to throw in a `a11y-only` `span` element

models/customer.js
jefflembeck: I totally agree. `getById` maybe? I named it after the `license-api` function, but it probably should be refactored.

handlers/org.js
jefflembeck: I agree with that. I originally started with shooting for the customer's id (from the Customer table in the `license-api`), but I figured out that it could take email and that required one less call.

assets/styles/org.styl
jefflembeck: Yes. Absolutely.

models/customer.js
rockbot: I like `getById` - it's super clear what it's for :-)

Another option would be to change the current `get` to `getCustomerStripeData` or something, since it's ever so slightly different

templates/org/info.hbs
rockbot: I guess my question is more about why it's a button/`input type=submit` within a tab; what will it lead to? is that where the 3rd party billing information comes in?

jefflembeck: Ah, yes. For our stop-gap 3rd party, this button triggers it.

PR #1136, opened 2015-08-10
Format all the things with esformatter

aredridel:


aredridel:
We could add that as a separate but matching set of rules -- integrate eslint in addition to esformatter. They work in similar ways, so there won't be a lot of impedance mismatch between them.

I prefer to have my editor auto-run the formatter when saving, though, rather than telling me about any perceived errors. Make it stop being a concern rather than highlighting it as a primary concern.

aredridel:
(TL;DR: I want to know about design errors and to stop thinking about formatting.)

jefflembeck:
> I prefer to have my editor auto-run the formatter when saving, though,

I save so frequently that this is fine with me.

jefflembeck:
Big questions left for me:

* Is this something we want to enforce/change at build time?
* Is there a config file that we could point everything to so the rules are deliberately set and we can change them if we want (see: https://github.com/millermedeiros/esformatter/blob/master/doc/config.md)?

aredridel:
I'm happy waiting to see if we need to enforce it at build time, or better, in a precommit hook. 

There is a config file -- you can put rules in package.json.

jefflembeck:
Oh that's perfect. :+1: Let's bikeshed rules that might go away from the default in as we go then? (wondering if we should pick double vs single quotes, for example)

aredridel:
I tend to accept the default unless there's a problem I'm solving. But let's bikeshed ongoing if we find something we care about.

PR #1139, opened 2015-08-11
Handle errors

aks-:


aredridel:
And in fact, the build failed because the lack of return on that callback made a callback called twice error.

aks-:
yeah, I made that change but to other branch :P Well, after adding return in front of callback I still get the failing tests. I will see later what's happening with it

models/user.js
```diff
@@ -242,7 +245,11 @@ User.prototype.getPackages = function(name, page, callback) {
 
       return resolve(body);
     });
-  }).nodeify(callback);
+  }).nodeify(callback)
+  .catch(function(err) {
+    console.log(err);
+    return Promise.reject(err);
```
aredridel: instead of `Promise.reject(err)` you can just `throw err`.

aredridel: And I still think this belongs in the calling code to handle errors, since it has context to deal with it appropriately. After the `.nodeify(callback)`, you're handling errors _from_ the callback, not just things related to this. You could move the catch to before the nodeify to not do that.

aks-: Oho I see. But if I do `throw err` in the catch. Isn't it same as not having catch?

Angelfirenze01:
Does this pull request include the problem with logging in while ad blockers are active?

aredridel:
It does not. This is just some refactoring to handle some errors.

PR #1142, opened 2015-08-12
Remove duplication in models/user

vanesyan:


vanesyan:
Oh, it's okay :smiley:

PR #1145, opened 2015-08-13
Org marketing page

jefflembeck:
First pass at the org marketing page!


routes/public.js
```diff
@@ -60,7 +60,7 @@ var publicRoutes = [
     method: "GET",
     handler: function(request, reply) {
       if (request.features.org_billing) {
-        return require('../facets/user/show-orgs')(request, reply);
+        return reply.view("org/index");
```
rockbot: ooh, excellent!

aredridel:
Doesn't matter too much whether i18n is now or later. It's similar work in either case.

PR #1158, opened 2015-08-14
Digest

rockbot:
this PR allows us to have auto-generated compiled css and js files. These files prevent the need to manually clear the cache in fastly whenever we make a style or client-side script change.


aredridel:
Wow. That is not how I'd have done it, but it's remarkably tidy! Nice find with rev-all there. It's an approach I hadn't really considered.

:sheep: it!

jefflembeck:
:sheep: !!

PR #1159, opened 2015-08-14
Adds a negative margin to the constructor.io ad

howonlee:
We are changing how the constructor-autocomplete widget works, so this will be necessary to get the ad to line up properly after.

Without the PR, with our change to the widget:
without_pr

With the PR, with our change to the widget:
with_pr


howonlee:
Apparently, Dan wants to add different things: sorry, we didn't coordinate properly on this one.

We will have a slightly bigger PR for you later

PR #1162, opened 2015-08-14
Adds a negative margin to the constructor.io ad (round 2)

howonlee:
We are changing how the constructor-autocomplete widget works, so this
will be necessary to get the ad to line up properly after.

Without the PR, with our change to the widget:
without_pr

With the PR, with our change to the widget:
with_pr

We made an abortive PR two hours ago, but this one is the actual PR

### DIFF
--- a/assets/styles/_header.styl
+++ b/assets/styles/_header.styl
@@ -285,6 +285,8 @@ nav
 
 #powered-by-constructor-io
   background-color greigh6
+  margin-left -8px
+  border-color greigh5
 
 #user-info
   position absolute


howonlee:
So that is the current version we're serving from the cdn. However, the metacpan folks, who are also giving us an ad, have it so that the autocomplete suggestions scroll, and the ad is not visible immediately. (see it at [metacpan.org](https://metacpan.org/), pick the author search and type in something)

We have a fix, that's what the fix looks like currently with the css settings you have on our debug proxy.

howonlee:
yes, this is the case; the PR is to make npm.js compatible with the fix to make it look not broken, which we ship on our end.

howonlee:
OK, we talked it over on our end and we will turn off ads before pushing the CDN change so it won't look wonky.

danmcc:
Hi @rockbot.  Sorry for the confusion on this.  We turned the banner off on our side.  If you can go ahead and push these PR changes, then we'll test thoroughly on our end to make sure the banner looks ok before pushing the changes on our side.

danmcc:
Thanks @rockbot!  We just tested it in a bunch of browsers (with the banner still turned off), and all looks well, so feel free to push whenever you like.  We'll re-test privately with the banner before turning it back on.

Btw, with the addition of the package descriptions, it looks like a scrollbar showed up.  You can get rid of that by increasing the [maxWidth](https://github.com/npm/newww/blob/master/assets/scripts/vendor/constructor-io.js#L9) to something like 500.  (I can submit a separate PR for that if ya want.)

Also, feel free to edit the CSS of the descriptions as you'd like.  Right now, the default is:

.autocomplete-suggestion-description {
        font-size: 0.8em;
        display: block;
        margin-left: 10px;
        line-height: 1.0em;
 }

Thanks!

danmcc:
Hi @rockbot!  Just wanted to check in on this.  Anything we can help with?

danmcc:
Awesome, thanks!  You rock, @rockbot !

PR #1163, opened 2015-08-14
adds vendor.min.js to versioned, cache-busting workflow

rockbot:
when we update vendor scripts, they get concatenated into vendor.min.js. It turns out we're caching vendor.min.js as well, so this PR lets us avoid the manually-nuke-the-cache issue.


jefflembeck:
:sheep:

PR #1164, opened 2015-08-17
Small fixes

aredridel:
- needless use of __dirname
- add the dev env to the dev script
- add .swp files to .gitignore (hi, vim! Bye, vim!)
- instantiate Error with new, though it doesn't matter in V8, it does in JS in general.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "server.js",
   "scripts": {
     "coverage": "npm test -- --reporter html > test/coverage.html && open test/coverage.html",
-    "dev": "gulp dev",
+    "dev": "NODE_ENV=dev gulp dev",
     "dev-db": "node dev/go.js",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "server.js",
   "scripts": {
     "coverage": "npm test -- --reporter html > test/coverage.html && open test/coverage.html",
-    "dev": "gulp dev",
+    "dev": "NODE_ENV=dev gulp dev",
```
rockbot: i'm fine with this, but it should be noted that `NODE_ENV=dev` is in the gulp file itself as well... maybe it shouldn't be there?

aredridel: Oh that is curious! I actually had an error like something wasn't set to dev mode.

aredridel: I think either place is reasonable.

PR #1168, opened 2015-08-18
removes urls that aren't actually urls

rockbot:
in general, urls that execute javascript (for example) _should_ be caught by the CSP, but in case they aren't, this PR should add another level of protection.

### DIFF
--- a/presenters/package.js
+++ b/presenters/package.js
@@ -80,6 +80,14 @@ module.exports = function(pkg) {
     delete pkg.homepage;
   }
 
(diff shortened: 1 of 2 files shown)


othiym23:
I am a tiny bit surprised that an inline script doesn't count as a valid URL, but aside from that, LGTM.

aredridel:
What @othiym23 said. But the definition of URL that exclude JavaScript happens to be the right one for us.

aredridel:
:sheep:

PR #1173, opened 2015-08-19
normalize package data, and use URLs with known schemes for homepage and repositories

aredridel:
This brings in [`normalize-package-data`](https://github.com/npm/normalize-package-data#readme) and then picks specifically acceptable things out of what it presents.

### DIFF
--- a/presenters/package.js
+++ b/presenters/package.js
@@ -2,19 +2,20 @@ var _ = require('lodash'),
   cache = require('../lib/cache'),
   fmt = require('util').format,
   gh = require('github-url-to-object'),
-  isUrl = require('is-url'),
   marky = require('marky-markdown'),
   metrics = require('../adapters/metrics')(),
   normalizeLicenseData = require('normalize-license-data'),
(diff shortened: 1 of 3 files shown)


presenters/package.js
```diff
@@ -2,19 +2,20 @@ var _ = require('lodash'),
...
   presentCollaborator = require("./collaborator"),
   presentUser = require("./user"),
+  npd = require('normalize-package-data'),
   url = require('url');
 
 var MINUTES = 60; // seconds
 var CACHE_TTL = 5 * MINUTES;
 
 module.exports = function(pkg) {
+  npd(pkg);
```
iarna: This is a big win, since current `normalize-package-data` has `hosted-git-info`'s magic for intuiting urls from various git-as-as-service providers.

aredridel: Indeed! Also, I think long term, it'd be nice to extend `hosted-git-info` to include the related home page, to use to fill in the package home page, rather than having that relationship created elsewhere.

PR #1176, opened 2015-08-19
Split the signup handler by method

aredridel:


aredridel:
Ooh, good call.

aredridel:
Updated with that. Good change.

PR #1180, opened 2015-08-20
Missing payment form

rockbot:
what tests am I missing?

I have refactored some of the customer handler tests to clarify what's happening, and when. other than that, though, the only way to guarantee that certain functions are being called when we want them to be called is to use sinon spies. thoughts? ideas? halp?

### DIFF
--- a/handlers/customer.js
+++ b/handlers/customer.js
@@ -79,7 +79,7 @@ customer.updateBillingInfo = function(request, reply, callback) {
       }
 
       if (callback) {
-        return callback(err);
+        return callback(er);
       }
 
       return reply.redirect('/settings/billing?updated=1');
@@ -111,6 +111,10 @@ customer.subscribe = function(request, reply) {
   var planInfo = {
     plan: plans[planType]
   };
+
+  if (planType === 'orgs' && !request.features.org_billing) {
+    return reply.redirect('/settings/billing');
+  }
 
   if (request.features.org_billing && planType === 'orgs') {
     planInfo.npm_org = request.payload.orgName;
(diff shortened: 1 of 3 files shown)


handlers/customer.js
```diff
@@ -78,6 +78,10 @@ customer.updateBillingInfo = function(request, reply) {
         request.logger.error(er);
       }
 
+      if (callback) {
+        return callback(err);
```
aredridel: actually, this should be `er`, not `err` -- we're nested here and the names are different.

aredridel:
There's no test for when the feature is disabled or any of those paths.

aredridel:
Otherwise, reasonable coverage, I think, at least of the obvious non-exceptional conditions.

handlers/customer.js
```diff
@@ -154,18 +159,21 @@ customer.subscribe = function(request, reply) {
...
   } else {
-    request.customer.createSubscription(planInfo, function(err, subscriptions) {
-      if (err) {
-        request.logger.error("unable to update subscription to " + planInfo.plan);
-        request.logger.error(err);
-      }
+    customer.updateBillingInfo(request, reply, function(err) {
+      request.customer.createSubscription(planInfo, function(err, subscriptions) {
+        if (err) {
+          request.logger.error("unable to update subscription to " + planInfo.plan);
```
aredridel: Can this be multiple args instead of + ? That way if planInfo.plan is a non-string it still outputs reasonably?

aredridel:
This looks great!

aredridel:
:sheep:

PR #1182, opened 2015-08-21
Support deep linking for tabs

### LINKED ISSUE #1154: Org management - Tabs aren't sticky
The tabs should update the url and should be immediately accessible upon refresh

jefflembeck:
Make sure it stays at the top of the page when you click the tab nav.
Make sure the back button is supported.
Fix #1154


aredridel:
Shiny. Not a bad approach!

PR #1188, opened 2015-08-25
Price area styling

rockbot:
I have [bent CSS to my will](https://twitter.com/rockbot/status/635963757638750208).

screen shot 2015-08-25 at 12 05 43 pm

### DIFF
--- a/assets/styles/org.styl
+++ b/assets/styles/org.styl
@@ -144,6 +144,8 @@
   vertical-align middle
   width 33%
   margin-left -4px
+  height 400px
+  line-height 1.5
 
   &:first-child
     margin-left 0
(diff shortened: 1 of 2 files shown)


assets/styles/org.styl
```diff
@@ -161,21 +162,46 @@
...
+  padding-bottom 5px
+
 .org-pricing .org-price .price
   color npmred
   font-size 31px
   display block
   font-weight 600
 
+  &:before
+    content '$'
```
aredridel: Good for now, but we may want to move that into a presentation layer if we have to deal with currencies.

aredridel:
Shiny!

jefflembeck:
:+1:

assets/styles/org.styl
rockbot: totally agreed

PR #1235, opened 2015-09-01
Separate startup sequence from server.js for future re-use

aredridel:


aredridel:
Let's merge this and other heave-ho to the tests after we get P0 out.

PR #1236, opened 2015-09-01
Billing page

jefflembeck:
The billing page has been updated from every angle possible - it only shows active orgs. It allows you to unsubscribe (without deleting), it shows your last date that the organization is going to be active, it blocks you from accessing orgs that you're not subscribed to.

It has also been updated from the aspect of appearance to better match the design.

### DIFF
--- a/models/customer.js
+++ b/models/customer.js
@@ -167,6 +167,16 @@ Customer.prototype.createSubscription = function(planInfo, callback) {
   });
 };
 
+Customer.prototype.cancelSubscription = function(subscriptionId, callback) {
+  var url = this.host + '/customer/' + this.name + '/stripe/subscription/' + subscriptionId;
+  Request.del({
+    url: url,
+    json: true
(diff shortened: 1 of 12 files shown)


aredridel:
Not bad. A couple ignored errors I'd love to see something more definite done with in the future, but I see no big structural flaws in this!

models/customer.js
```diff
@@ -167,6 +167,16 @@ Customer.prototype.createSubscription = function(planInfo, callback) {
...
+Customer.prototype.cancelSubscription = function(subscriptionId, callback) {
+  var url = this.host + '/customer/' + this.name + '/stripe/subscription/' + subscriptionId;
+  Request.del({
+    url: url,
+    json: true
+  }, function(err, resp, body) {
+    callback(err, body);
+  });
+};
+
```
rockbot: can you throw together a really quick test for this new method?

PR #1237, opened 2015-09-01
Org Creation

rockbot:
This PR is for handling the `/org/create` route - whether you're paid or not.

There is almost certainly stuff that's missing in here; I wanted to get this PR made, though, so that we can start working through all the pieces as a team.

### DIFF
--- a/assets/styles/forms.styl
+++ b/assets/styles/forms.styl
@@ -121,7 +121,7 @@ form p
 
   a
     text-decoration none
-    margin 0 5px
+    // margin 0 5px
 
     &:hover
       color greigh2
--- a/assets/styles/forms.styl
+++ b/assets/styles/forms.styl
@@ -121,7 +121,6 @@ form p
 
   a
     text-decoration none
-    // margin 0 5px
 
     &:hover
       color greigh2
--- a/handlers/customer.js
+++ b/handlers/customer.js
@@ -146,7 +146,7 @@ customer.subscribe = function(request, reply) {
                   return reply.view('errors/internal', err);
                 }
 
-                return reply.redirect('/settings/billing', opts);
+                return reply.redirect('/org/' + planInfo.npm_org, opts);
               });
 
           });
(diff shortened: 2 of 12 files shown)


assets/styles/forms.styl
```diff
@@ -121,7 +121,7 @@ form p
 
   a
     text-decoration none
-    margin 0 5px
+    // margin 0 5px
```
aredridel: Why'd this lose its margin, without just deleting it?

handlers/customer.js
```diff
@@ -105,76 +104,121 @@ var plans = {
...
-                return reply.redirect('/settings/billing', opts);
-              });
+        Org(request.loggedInUser.name)
+          .get(planInfo.npm_org).then(function() {
+          var err = new Error("Org already exists");
+          err.isUserError = true;
+          throw err;
+        }).catch(function(err) {
+          if (err.statusCode !== 404) {
+            throw err;
```
aredridel: Is something catching these errors?

handlers/customer.js
```diff
@@ -105,76 +104,121 @@ var plans = {
...
+          if (err) {
+            request.logger.error("unable to update subscription to " + planInfo.plan);
+            request.logger.error(err);
+          }
+
+          if (typeof subscriptions === 'string') {
+            request.logger.info("created subscription: ", planInfo);
+          }
+
+          return reply.redirect('/settings/billing?updated=1');
```
aredridel: We probably want to integrate notifications here at some point; perhaps after the notification module is factored to separate 'collect errors' from 'store notices'

aredridel:
Wow, this looks great. Looks like it might need a `npm install verror --save` though

assets/styles/forms.styl
rockbot: good catch - forgot to delete (but also wanted to make sure i didn't jenga-fall the entire website for the one change without full smoke tests). deleting now!

handlers/customer.js
rockbot: :+1:

aredridel:
:sheep: it!

PR #1244, opened 2015-09-01
New profile sidebar

jefflembeck:
Slight design change implemented and create org link added.

### DIFF
--- a/templates/user/profile.hbs
+++ b/templates/user/profile.hbs
@@ -71,7 +71,6 @@
     {{> email-verify-nag}}
   {{/if}}
 
-  {{#unless isSelf}}
   {{#if resource.fullname}}
     <h2 class="fullname">{{resource.fullname}}</h2>
   {{/if}}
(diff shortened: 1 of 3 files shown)


templates/user/profile.hbs
```diff
@@ -56,25 +56,27 @@
...
+      {{/if}}
       <li><a href="/profile-edit">update your profile</a></li>
-      <li><a href="/password">change your password</a></li>
-      <li><a href="http://gravatar.com/emails/">set your avatar</a></li>
     </ul>
 
     {{> email-verify-nag}}
   {{/if}}
 
+  {{#unless isSelf}}
```
rockbot: when logged in, I see: 

screen shot 2015-09-01 at 5 08 16 pm

but when I look at another user's profile, I see:

screen shot 2015-09-01 at 5 08 40 pm

I like being able to see the stuff other people see on my profile (i.e. links for how to contact me), so that I know that when I change something in my profile, it's rendered correctly.

Think we could remove this line as a result?

jefflembeck:
Removed, :sheep:in

PR #1245, opened 2015-09-02
enables a /org?join-beta link for invited users (fixes #1205)

### LINKED ISSUE #1205: Notify invited users to sign in so that they can access the org stuff
For phase 0, we're going to be inviting teams to try orgs via an email link, i.e. `/org?hereismyinvitecode`. When a user goes to that link, we should encourage them to login, instead of shipping them to the "register your interest in orgs" form.

rockbot:


aredridel:
Shiny! :sheep:

jefflembeck:
Thiiis looks awesome.

PR #1246, opened 2015-09-02
Remove test for feature_page flag that isn't even

aredridel:


aredridel:
Yeah, that was my guess, but I can find no reference and I'd have to go spelunking to figure it out.

aredridel:
Okay, couldn't resist. It was `process.env.FEATURE_ACCESS_PAGE` -- but that's been removed since.

PR #1247, opened 2015-09-02
Use tap to run tests

aredridel:
Todo: Speed up initialization of tests


facets/company/show-send-contact.js
```diff
@@ -1,11 +1,11 @@
...
-});
-
 module.exports = function showSendContact(request, reply) {
+  var zdClient = require('node-zendesk').createClient({
+    username: process.env.ZENDESK_USERNAME,
+    token: process.env.ZENDESK_TOKEN,
+    remoteUri: process.env.ZENDESK_URI,
+    disableGlobalState: true
+  });
+
```
rockbot: why declare this inside the function instead of outside?

presenters/package.js
```diff
@@ -139,6 +138,7 @@ function processReadme(pkg) {
   }
 
   var cacheKey = pkg.name + '_readme';
+  var marky = require('marky-markdown');
```
rockbot: similar to above, why not have it declared once at the top vs twice closer to where it's being used?

(genuinely curious :-))

routes/public.js
```diff
@@ -103,6 +106,7 @@ var publicRoutes = [
   }, {
     path: "/whoshiring",
     method: "GET",
+    x: console.log('rba', Date.now() - xx),
```
rockbot: what does this `x` do?

test/adapters/bonbon.js
```diff
@@ -260,3 +260,9 @@ describe("bonbon", function() {
   });
 
 });
+
+Lab.report([lab], {
+  reporter: 'tap',
+  output: process.stdout,
+  leaks: false
+});
```
rockbot: this is a stopgap until we transition the tests over to native tap, yes?

presenters/package.js
aredridel: It delays the initialization so that it's not incurred at startup time -- which greatly speeds tests and accessing requests that aren't touching these paths.

See the commit  5da7c93 -- `Reduce startup time by making marky-markdown load lazy`

routes/public.js
aredridel: That's from my temp to be removed timing-information commit.

test/adapters/bonbon.js
aredridel: Yep!

PR #1252, opened 2015-09-02
fixes bug where already-paid user was unable to create an org

rockbot:
the issue was simple: since we don't show the credit card form when we already have billing information, we were sending an empty object to stripe, which caused a 400. this PR fixes it. :-)


aredridel:
Fair! 

It's a bit indirect -- far from the actual thing that it relates to, can't spot the business logic in it -- but should work and it's the right change in context!

:sheep:

Issue #1257, opened 2015-09-02
Orgs: Link to create an org is broken


aredridel:
via @seldo: Also: after you verify your email address, the page links to "create an org" but the link is broken (/orgs instead of /org I think)

aredridel:
@seldo: I can't replicate this -- this was a non-org user without a validated email address? Logged in to start?

aredridel:
This is probably fixed by #1258 too.

mstade:
I tried following the link, and had forgotten my password. I reset it, followed the link again, and was redirected to my profile page. I made sure my email was verified, but this changed nothing. The [org link][1] just keeps redirecting to my profile page.

[1]: https://www.npmjs.com/org?join-beta

aredridel:
That is indeed happening for users who aren't eligible for the first round of beta! Investigating why now!

PR #1258, opened 2015-09-02
allows /orgs?join-beta to work as well (fixes #1250)

### LINKED ISSUE #1250: Orgs: /org?join-beta is different than /orgs?join-beta
Both should go to to the login page.

rockbot:
even though the `/orgs` redirect works, it doesn't carry the query parameters along, so we have to be explicit about it.

also: added some tests to ensure no regressions later on.

### DIFF
--- a/routes/public.js
+++ b/routes/public.js
@@ -56,6 +56,20 @@ var publicRoutes = [
       return reply.redirect("/private-modules").code(301);
     }
   }, {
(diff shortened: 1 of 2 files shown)


aredridel:
Works for me. Bet it also fixes #1257 though I can't test that one without more info

aredridel:
:sheep:

PR #1260, opened 2015-09-02
Add regex: feature flag matches

aredridel:


lib/feature-flags.js
```diff
@@ -52,7 +55,10 @@ var isFeatureEnabled = module.exports = function isFeatureEnabled(feature, reque
   }
 
   if (Array.isArray(value) && request && request.auth && request.auth.credentials) {
-    return value.indexOf(request.auth.credentials.name) > -1;
+    var name = request.auth.credentials.name;
+    return value.some(function(potentialUser) {
+      return potentialUser === request.auth.credentials.name || potentialUser.test && potentialUser.test(name);
+    });
```
rockbot: this is magical. that is all.

PR #1265, opened 2015-09-03
adds list of user's orgs to profile page if they are logged in

rockbot:
Note: this only works if you are logged in, are looking at your profile page, and have created organizations. If you have no orgs, are not logged in, or are looking at someone else's profile page, you will *not* see any org lists.

screen shot 2015-09-02 at 4 21 04 pm


aredridel:
:sheep: !

Issue #1253, opened 2015-09-02
Orgs: Adding non-existing user causes internal server error


aredridel:
if the user doesn't exist, send back a user-understandable error instead of a 500

PR #1266, opened 2015-09-03
Add user exception handling

jefflembeck:
This makes it so we don't have a major crash if a user doesn't exist while adding to an org.


jefflembeck:
Fixes #1253

aredridel:
Works for me! :sheep:

PR #1270, opened 2015-09-03
Add a feature flag to bypass email verification

aredridel:


handlers/user.js
```diff
@@ -97,6 +98,22 @@ exports.handleSignup = function signup(request, reply) {
...
           request.logger.info('created new user ' + user.name);
 
+          var almostARequest = {
+            auth: {
+              credentials: {
+                name: user.name
+              }
+            }
+          };
+          if (feature('bypass_email_verify', almostARequest)) {
```
rockbot: so we'd put a line like `FEATURE_BYPASS_EMAIL_VERIFY=true` in the .env? how would one confirm the contents of the email?

aredridel: We'd confirm the contents of the email... the old fashioned way. No change here, and while it's worth testing that's not a part of the site under active development.

jefflembeck:
makes sense in my head :sheep:

PR #1279, opened 2015-09-04
removes mention of orgs to new users who are not in the beta

rockbot:


templates/user/billing.hbs
```diff
@@ -1,6 +1,6 @@
 {{> profile-secondary-nav}}
 
-<div class="container">
+<div class="container {{#unless features.org_billing}}narrow{{/unless}}">
```
aredridel: I'd love to refactor this or at least mark it for clean-up once we hit GA, but should work great.

aredridel:
:sheep:

templates/user/billing.hbs
rockbot: indeed! at least `features.org_billing` is easy to site-wide search/replace when we get to GA :-D

PR #1280, opened 2015-09-04
Make clear button be accessible on small screens

alrra:
See: #1074.

--

Cc: @aredridel.


aredridel:
Wow, that looks great.

:sheep:

PR #1282, opened 2015-09-04
Filter out bad subs

rockbot:
handles an issue where sometimes the license API returns sub-optimal data when a user tries to sign up for private modules after already being a paid member of an org.


models/customer.js
```diff
@@ -21,7 +21,6 @@ var Customer = module.exports = function(name, opts) {
 };
 
 Customer.prototype.getById = function(id, callback) {
-  var self = this;
```
aredridel: Nice! Lost binding!

aredridel:
V. nice.

PR #1286, opened 2015-09-04
Potential fix for the LastPass problem

dancrumb:
So, this is a potential fix for the problem whereby LastPass is populating the honeypot field.

It requires LastPass to be appropriately configured (https://forums.lastpass.com/viewtopic.php?f=6&t=78817), but at least it shows goodwill on your part


aredridel:
I think this would work!

jefflembeck:
seems good to me :sheep:

PR #1287, opened 2015-09-04
copy change to group-creation page

cowperthwait:


templates/org/create.hbs
```diff
@@ -25,11 +25,10 @@
           <p class="description">Your scope becomes part of your package names and <strong class="fake-strong">cannot</strong> be changed.</p>
 
           <fieldset>
-            <legend>organization type</legend>
```
rockbot: I believe this is for accessibility reasons; it doesn't actually render on the page.

cc: @jefflembeck

cowperthwait: it renders on the page: 
http://dl.dropbox.com/u/73197268/Screenshots/htnd.png

and now that there is only one organization to choose, the whole flow is pretty superfluous, but at a minimum we could remove the legend text instructing us which (of one) radio button to select

jefflembeck: Yeah, it's superfluous and not necessary at this time.

PR #1298, opened 2015-09-04
StripeCheckout is not available at load and that's okay

seldo:


aredridel:
Shiny. Looks right!

jefflembeck:
Got pulled in locally

PR #1309, opened 2015-09-09
renaming npmE to npm on-site

bcoe:
We need to land these changes to the email copy at the same time:

https://github.com/npm/email-templates/pull/14

I've already published the updated installer:

https://www.npmjs.com/package/npmo


cowperthwait:
lgtm

PR #1310, opened 2015-09-09
409 return status isn't ready yet in production like it is in staging

jefflembeck:
This provides us with a handler for the 500 for the time being.


aredridel:
WFM pending travis

PR #1313, opened 2015-09-10
Error handling

jefflembeck:
Wider error handling for org agent and customer model.


aredridel:
this looks complete. I spent some time reading this morning, and I think this covers all the cases.

PR #1315, opened 2015-09-10
If org user, sign-up for private modules should directly link to priv…

jefflembeck:
…ate module payment info


jefflembeck:
@rockbot it's feature flag based, so, unfortunately, manually - if the feature flag is turned on, that link will go directly to the payment info tab

PR #1318, opened 2015-09-10
sends 409s back to org/create with user-friendly messaging

rockbot:
am I missing anything?

### DIFF
--- a/agents/org.js
+++ b/agents/org.js
@@ -43,7 +43,7 @@ Org.prototype.create = function(name, callback) {
       }
 
       if (resp.statusCode >= 400) {
-        err = new Error(body.message);
+        err = new Error(body);
         err.statusCode = resp.statusCode;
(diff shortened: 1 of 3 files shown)


agents/org.js
```diff
@@ -43,7 +43,7 @@ Org.prototype.create = function(name, callback) {
       }
 
       if (resp.statusCode >= 400) {
-        err = new Error(body);
+        err = new Error(body.message);
         err.statusCode = resp.statusCode;
```
aredridel: The change to the api for body.message was reverted, I think

aredridel:
Just one question!

agents/org.js
rockbot: it's what's on staging right now... cc @chrisdickinson

aredridel:
:sheep: :sheep: :sheep:

PR #1324, opened 2015-09-14
Paid/unpaid

jefflembeck:
Switching back to using switches for payment status of a user in an org. Only now, they work!


aredridel:
Shiny. The action to switch works?

aredridel:
Interesting failure in the tests there.

jefflembeck:
Yeah. Fixing those up. Less calls so mocking issues.

aredridel:
:sheep

PR #1327, opened 2015-09-15
Marky fork - oniguruma memory leak fixes

soldair:
adding @soldair/marky-markdown fork with the oniguruma memory leak fix https://github.com/atom/node-oniguruma/pull/36

don't want to publish an offical marky-markdown release that depends on my fork of highlights, first-mate, and oniguruma so for now we can depend on the fork


soldair:
I'm working on a support for node 10 in the oniguruma fork.

soldair:
ok all is green on stage. merge?

PR #1329, opened 2015-09-15
Hide downloads on scoped packages

### LINKED ISSUE #1328: Public scoped package download counts show zero
per @othiym23, "package counts haven't yet been implemented for scoped packages, because we have to add ACL logic to filter out private packages"

This issue imo is either to a) do that ^, or b) hide them (preferably with a note explaining the deficit) until they're implemented, since zero isn't accurate.

aredridel:
Fixes #1328

### DIFF
--- a/handlers/package.js
+++ b/handlers/package.js
@@ -33,7 +33,9 @@ exports.show = function(request, reply) {
     .then(function(results) {
       var pkg = results.package;
       pkg.dependents = results.dependents;
-      pkg.downloads = results.downloads;
+      if (pkg.name[0] != '@') {
+        pkg.downloads = results.downloads;
+      }
 
       if (pkg && pkg.time && pkg.time.unpublished) {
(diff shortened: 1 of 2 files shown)


aredridel:
Yeah, that was my thought too. First, stop lying, then make it right.

handlers/package.js
```diff
@@ -53,6 +55,8 @@ exports.show = function(request, reply) {
         && (typeof pkg.collaborators === "object")
         && (loggedInUser.name in pkg.collaborators);
 
+      pkg.hasStats = pkg.downloads || (pkg.bugs && pkg.bugs.url) || (pkg.pull_requests && pkg.pull_requests.url);
```
ljharb: won't this hide stats for any package with zero downloads?

aredridel: correct. But for public packages, that's ~never a thing, thanks to mirrors and analysis tools and such.

PR #1331, opened 2015-09-15
Paid unpaid redux

jefflembeck:
Make that thing work with ajax.

Error-handling needs to be added still. Let's do that as part of this, just wanted to keep it up to date :smile:


assets/scripts/switch-submission.js
```diff
@@ -0,0 +1,29 @@
...
+      .fail(function() {})
+  }.bind(this));
+};
+
+module.exports = function() {
+  $(function() {
+    var checkboxes = $(".switch");
+
+    $.each(checkboxes, function(idx, checkbox) {
+      var input = new Switch(checkbox);
```
aredridel: Making a whole class, just to instantiate it once, and call a single method on it seems kinda indirect. What's the advantage of this approach?

jefflembeck: Some of this is certainly YAGNI, but the structure allows for some easy tracking of state and potential growth if we decide to play with growth. Also, I find it a bit easier to unit test.

aredridel: Fair enough!

I've usually taken the approach to put something like this bound to a parent element, and have only one instance no matter how many switches, but this should work nicely.

jefflembeck: Yeah for sure. I think that I'm mostly doing this multiple binding because of the amount of forms. That being said, I could put the listener on the table, listen for the target and grab the specific form on that.

I'll think of a way to test that well

jefflembeck:
Last step: make sure all tables reflect updated data.

PR #1332, opened 2015-09-15
homepage: move 'getting started' above the fold

isaacs:
This patch moves the "getting started" bits from the bottom to just below the "things people install a lot" bit, and reduces the logo bunch so that they take up a lot less space.

I'm a bit out of practice with doing frontend stuff, and especially there's probably some stylus stuff that should be done differently.  But it seems like this sorta works?  I guess?

Review welcome, and deployment not at all urgent.  This is just something that's been bugging me for a while :)

### DIFF
--- a/assets/styles/layout.styl
+++ b/assets/styles/layout.styl
@@ -169,31 +169,29 @@ ul.columnar
       @extend .red-on-hover
       font-weight 400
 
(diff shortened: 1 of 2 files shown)


aredridel:
:+1:

PR #1334, opened 2015-09-17
updates the codebase to hapi 9.x

rockbot:
a few API changes, as per the [9.0.0 Release Notes](https://github.com/hapijs/hapi/issues/2682). This PR will set us up for [hapi 10.x](https://github.com/hapijs/hapi/issues/2764), which has zero API changes, but will allow us to transition to Node 4.

if I've learned anything from this PR, it's that I look forward to the day when we no longer need `mocks/server.js`, which is a huge pain and leaves `server.js` completely untested.

### DIFF
--- a/adapters/plugins.js
+++ b/adapters/plugins.js
@@ -8,6 +8,8 @@ module.exports = [
     }
   },
   require('scooter'),
+  require('inert'),
+  require('vision'),
   {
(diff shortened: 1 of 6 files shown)


aredridel:
Wow, that's easy! Nothing like the port from 7 to 8.

:sheep:

PR #1340, opened 2015-09-18
Http response splitting

jefflembeck:
Work around http response splitting bug that occurs in node > v0.10

### DIFF
--- a/routes/public.js
+++ b/routes/public.js
@@ -2,6 +2,7 @@ var fmt = require('util').format;
 var fs = require('fs');
 var _ = require('lodash');
 var async = require('async');
+var validate = require('validate-npm-package-name');
 
 var unathenticatedRouteConfig = {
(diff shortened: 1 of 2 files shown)


routes/public.js
```diff
@@ -179,11 +180,12 @@ var publicRoutes = [
     ],
     method: "GET",
     handler: function(request, reply) {
-      if (request.params.scope) {
-        return reply.redirect("/package/" + request.params.scope + "/" + request.params.project).code(301);
-      } else {
-        return reply.redirect("/package/" + request.params.package).code(301);
+      var fullPackageName = request.params.package || (request.params.scope + '/' + request.params.project);
+      if (!validate(fullPackageName).validForOldPackages) {
```
ceejbot: Heh. I like this approach: very specific to what the user was trying to do.

PR #1351, opened 2015-09-28
we don't use the npm2es depdendency in the www codebase

bcoe:
as far as I can tell, we don't use the npm2es dependency in newww (this service is deployed separately, and has its own operational concerns). The git dependency causes issues for the Alpine server I'm building newww on (CC: @aredridel).


aredridel:
:sheep:

PR #1354, opened 2015-09-28
a couple fixes to how we package up source for the website

bcoe:
* `dotenv` should not be a dev-dependency (we need it to run the production site).
* we should publish the static assets generated by gulp.


aredridel:
shiny!

PR #1356, opened 2015-09-30
Validation of url

jefflembeck:
URLs that have redirects with user input must be validated.


aredridel:
:sheep:

aredridel:
we should so get the mitigation for this in the abstract into hapi :/

PR #1360, opened 2015-09-30
Remove download counts from npmowww

aredridel:


bcoe:
Looks good to me, the plan is to simply eliminate download counts entirely -- I think this is a good call until we have our own internal metrics we can track.

PR #1363, opened 2015-10-02
fills in missing form actions (fixes #1358)

### LINKED ISSUE #1358: Users accessing the site from MS Edge are unable to log in
Over the last week or so, we've received a handful of support messages from users attempting to log in to the website using MS Edge and failing. 

Despite the use of proper credentials or new credentials via a password reset, when entered, the user is redirected to a 404 with a `https://www.npmjs.com ` url. 

edge

If a VM is needed for testing, one can be launched using [Microsoft Edge - Dev](https://dev.modern.ie/tools/vms/windows/) site.

rockbot:
Microsoft Edge (their latest browser) doesn't like empty `action` attributes on forms, which was why users were unable to login or signup on our website.

This PR fixes that (and other missing action attributes) :-)

### DIFF
--- a/templates/enterprise/find-license.hbs
+++ b/templates/enterprise/find-license.hbs
@@ -10,7 +10,7 @@
     <p>Thanks for choosing to license npm On-Site. To get started, enter the email address you
       wish to use for billing.</p>
 
-    <form method="post">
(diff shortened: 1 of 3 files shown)


aredridel:
Nice! :sheep:

PR #1366, opened 2015-10-05
Excise user profiles for npmo

aredridel:


routes/public.js
```diff
@@ -276,22 +278,27 @@ var publicRoutes = [
...
     handler: require('../facets/user/show-profile')
   }, {
     path: "/profile/{name}/packages",
     method: "GET",
+    feature: "!npmo",
     handler: require('../facets/user/show-packages')
   }, {
     path: "/signup",
     method: "GET",
+    feature: "!npmo",
```
bcoe: this is a slick way to toggle routes.

templates/package/show.hbs
```diff
@@ -102,17 +109,25 @@
     <ul class="humans collaborators">
       {{#each collaborators}}
         <li>
-          <a href="/~{{name}}" rel="tipsy" title="{{name}}">
-            {{#if avatar.medium}}<img src="{{avatar.medium}}" alt="{{name}}">{{/if}}
-          </a>
+          {{#unless ../../../features.npmo}}
+            <a href="/~{{name}}" rel="tipsy" title="{{name}}">
```
bcoe: simply disabling the link since we can't facilitate profile pages yet? seems reasonable to me.  The avatar/publisher are provided by registry relational?

bcoe:
:shipit: my only questions are out of curiosity.

routes/public.js
rockbot: where does the `feature` get picked up? (is it in another PR that's already been merged in? apologies if i've missed it!)

aredridel: It is already merged in -- in `adapters/bonbon.js` the feature flags are processed into the request, and in the route setup code in `routes/index.js`, there is an explicit filter by feature flag.

templates/package/show.hbs
aredridel: Publisher is -- avatar may not be and I'll have to check. (We also depend on gravatar in newww, so it's not obvious.)

PR #1369, opened 2015-10-07
Agentize collaborator model

rockbot:
bye bye `.new()`, hello agent!

PR #1371, opened 2015-10-07
Agentize downloads model

rockbot:
another!

one difference of note between this PR and #1369: in this one I'm doing the logic for the bearer token within the agent itself; in the other I'm doing it in the handler.

Thoughts on which is better? I'm kind of leaning towards doing it in the agent, but would be curious about others' thoughts... :-)


aredridel:
Hm. I don't have a strong feeling either way. I kinda like the other one more since it more cleanly separates 'web' concerns from the lower level API. I'm imagining using this in a non-web context like a command line tool and making magic objects that are the way they are only to satisfy what it looks for.

jefflembeck:
I actually prefer the other way as well - I like the idea of separating the request from the agent itself.

PR #1372, opened 2015-10-07
Agentize customer model

rockbot:
Do we even care about Customer.host? I've found a way to (sort of) keep it, but I don't think it's actually useful... thoughts?


aredridel:
I don't think we need to keep `Customer.host`. This looks great as is though! :sheep: it!

PR #1373, opened 2015-10-07
Agentize package model

rockbot:
Got rid of `this.host` entirely here... yeah, I don't think I miss it at all.


aredridel:
:sheep: SO nice to see this happening!

PR #1375, opened 2015-10-08
User org migration

jefflembeck:
This is adding the user to org migration feature. This is strictly functionality at this point and needs not only intense code review, but also appropriate styling.


handlers/customer.js
```diff
@@ -236,18 +243,32 @@ customer.subscribe = function(request, reply) {
           throw err;
         }
 
-        // org doesn't yet exist
-        return Org(request.loggedInUser.name)
-          .create(planInfo.npm_org)
+        // org doesn't yet exist, transfer user then create org
+        var start = newUser ? User.new(request).toOrg(loggedInUser, newUser) : P.resolve(null);
```
rockbot: what will the `P.resolve(null)` do here? since we're not returning, will it do what you're hoping it will do?

aredridel: It's a clever trick for getting a promise to chain from that does nothing. "If we need this stuff, do it and chain from there, else just ... give me a resolved promise to then from"

handlers/customer.js
```diff
@@ -236,18 +243,32 @@ customer.subscribe = function(request, reply) {
...
               .then(function(subscription) {
                 if (typeof subscription === 'string') {
                   request.logger.info("created subscription: ", planInfo);
                 }
-                return request.customer.extendSponsorship(subscription.license_id, request.loggedInUser.name);
+                return Customer(loggedInUser).extendSponsorship(subscription.license_id, loggedInUser);
               }).then(function(extendedSponsorship) {
-              return request.customer.acceptSponsorship(extendedSponsorship.verification_key);
+              return Customer(loggedInUser).acceptSponsorship(extendedSponsorship.verification_key);
             }).then(function() {
```
rockbot: at some point we need to reset request.customer with the new customer data (since the customer functions refer to `this.name`)

aredridel: Hmm. Can we redirect the user to a page, so we get a clean request, with it filled in properly from primary sources?

jefflembeck: We can, but I struggle with creation of sponsorship/extend/accept in a `GET`

jefflembeck: I think that might make the workflow more complicated if we add another step.

aredridel: Just a redirect to get to the 'it worked, you're done converting' screen maybe? After the final action?

That way the inteterminate half-changed state is at least isolated to the orgs code?

routes/authenticated.js
```diff
@@ -111,9 +112,54 @@ module.exports = [
...
+      return reply.view('org/billing', {
+        fullname: request.query.fullname,
+        orgScope: request.query.orgScope,
+        newUser: request.query['new-user'],
         stripePublicKey: process.env.STRIPE_PUBLIC_KEY
+
       });
+
+
+
```
rockbot: would it make sense to put these into their own handler files at this point? i.e. make `routes/authenticated.js` more of a lookup type of file?

templates/org/transfer.hbs
```diff
@@ -0,0 +1,38 @@
...
+        <a class="btn" href="/org/create">go back</a>
+        </section>
+      </form>
+  </div>
+  <div class="sidebar">
+    <div class="inset bottom-img">
+      <h3>Transferring your username</h3>
+      <ul>
+        <li><span>No one will be able to log in as @{{orgScope}} anymore, as it will be an org scope, it has no password</span></li>
+        <li><span>All Orgs have a dedicated namespace, unlimited teams, and publicll the packages belonging to @{{orgScope}} the user will now belong to @{{orgScope}} the org
```
rockbot: "publicll"? publicly, perhaps?

templates/org/transfer.hbs
```diff
@@ -0,0 +1,38 @@
...
+        package hosting.</span></li>
+        <li><span>Your new username will have your existing email and password credentials</span></li>
+        <li><span>This cannot be undone</span></li>
```
rockbot: we should embolden this. make it super duper zomg clear.

test/handlers/org.js
```diff
@@ -238,6 +238,124 @@ describe('getting an org', function() {
...
+        method: "GET",
+        credentials: fixtures.users.bob
+      };
+
+      server.inject(options, function(resp) {
+        userMock.done();
+        orgMock.done();
+        expect(resp.statusCode).to.equal(302);
+        expect(resp.request.response.headers.location).to.match(/org\/create/);
+        done();
```
rockbot: this test is very similar to the one above it; maybe check to make sure that the error messages being put onto the page are specific to the type of error?

templates/org/transfer.hbs
ncawthon: FWIW - there is a copy deck in the TnO channel, that haz some em-biggened / em-boldings... Kasey approved.

test/models/user.js
```diff
@@ -655,4 +655,83 @@ describe("User", function() {
...
+        .reply(200, orgObj);
+
+      User.toOrg("npmjs", "npmjs-admin", function(err, data) {
+        userMock.done();
+        expect(err).to.be.null();
+        expect(data.name).to.equal("npmjs");
+        done();
+      });
+    });
+  });
```
rockbot: my only thought on this is: are there any security holes hiding here? We've got a bearer that doesn't match the new username making an irreversible change. I'm fairly certain this is fine (since we do it in loads of other places), but it's worth asking ourselves.

handlers/customer.js
rockbot: :+1:

routes/authenticated.js
aredridel: Definitely would be nice to reduce it down to a simple table.

templates/org/transfer.hbs
jefflembeck: Definitely

jefflembeck:
Tests are passing. Smoke tests are passing. Going to merge this in.

PR #1379, opened 2015-10-09
Add a place to show billing errors for orgs

aredridel:


aredridel:
It's all client-side, because stripe. We never actually receive credit card numbers.

PR #1381, opened 2015-10-12
remove dependentCount from fixtures since it's gone in the responses in live

aredridel:


jefflembeck:
:heart: DELETING THINGS :heart:

PR #1383, opened 2015-10-13
allow empty org to still have users added by the org admin

rockbot:
Before this PR: if a user deleted themselves from their org before adding a new member to the org, the user would be unable to add any new members to the org (because the table to do so would disappear)

After this PR: an org creator can delete themselves but continue adding others :-)


jefflembeck:
Good catch.

handlebars has a sweet shortcut for the `if each /each else /if` pattern with `each else /each`

aredridel:
Oh shiny, @jefflembeck

aredridel:
Works!

jefflembeck:
\o/ :sheep:

PR #1385, opened 2015-10-13
Rename packages people 'npm install' a lot to featured packages

ForbesLindesay:
Since this is a manually curated list that has not been updated in 8 months, it seems more appropriate to be transparent about its intended purpose.


aredridel:
I hear you.

It is in fact based on the metric of what's 'npm install'ed a lot, but manual and only periodically reviewed since logos need manual work.

Thanks for the PR, but we're going to leave it as is.

PR #1386, opened 2015-10-13
Scope check during signup

jefflembeck:
We need to check the scope used instead of just the username, for the case of an org name being used.


aredridel:
Utterly solid. :sheep:

PR #1387, opened 2015-10-13
allows feature flags to be set via user's resource

rockbot:
there's some repeated code in bonbon.js - L19-26 and L48-56 - but I'm not sure what the best way to DRY that up is... thoughts, @jefflembeck, @aredridel ?

In particular, do we keep some of that logic in adapters/bonbon.js or move it into lib/feature-flags.js?

### DIFF
--- a/adapters/bonbon.js
+++ b/adapters/bonbon.js
@@ -45,6 +45,15 @@ exports.register = function(server, options, next) {
         }
         if (user) {
           user.sid = request.auth.credentials.sid;
+          Object.keys(user.resource || [])
+            .filter(function(key) {
+              return key.match(/^feature_/i)
+            })
+            .forEach(function(key) {
+              if (user.resource[key] === 't') {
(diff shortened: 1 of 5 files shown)


aredridel:
I think moving it into `lib/feature-flags.js` is exactly the right approach, since with a reference to the request, it can get the user and do just that -- we can normalize and merge both sources of information and then check them together.

adapters/bonbon.js
```diff
@@ -45,6 +45,15 @@ exports.register = function(server, options, next) {
         }
         if (user) {
           user.sid = request.auth.credentials.sid;
+          Object.keys(user.resource || [])
```
chrisdickinson: I might recommend `|| {}` over `|| []` here to match the expected type of `user.resource`.

rockbot: :+1: good call!

aredridel:
This looks great! :sheep:

PR #1392, opened 2015-10-16
Adds tiny tonicdev integration to see if people like it

seldo:
https://tonicdev.com built a neat and fun npm integration. One day we might consider actually letting people test modules out inline on the package page. This lightweight link will let us test the water and see if people like Tonic and find it useful.


aredridel:
LGTM!

PR #1393, opened 2015-10-16
Fix next page bug of search

vramana:
@aredridel Please have a look :bow:


vramana:
On a side note, the 20 search results seems too long to scroll though for me. Wouldn't something like this be a little better 
npm-search

**Note** I removed the job advertisements. Probably can be added as a cell in the grid.

aredridel:
I'll let our designer know your thoughts!

@ncawthon -- a bit of input into search design!

aredridel:
This does indeed fix it! Thank you!

PR #1400, opened 2015-10-17
Normalize height of elements in <header>

vramana:
As shown in the below picture, #user-info, #npm-logo, #npm-search are at uneven distance from top due to uneven heights. 

npm-header-before

After this fix, all of them will be basically at the same hieght. The gist of fix is declare all the elements to have same height and instead of individually setting padding and margin on each element set large padding in wrapping element.

npm-header-after


vramana:
But the mobile view is got worse by this change. The header doesn't look so nice even before the change. This just adds a :cherries: : on top of it. **Before**:
1 - lurvqla

**After**:
2 - glseezf

I am thinking of a better solution than padding. Height seems to be the way to go. Might also work But I found an interesting thing. There is no way to logout once you login in. The change happened in this commit 332d6e18ed99d277114ec793a4d05b63dc08b6d1 I think this is not an intended behaviour. @aredridel Any thoughts?

aredridel:
Ooh, this looks like things I'd noticed in the making the top bar flexbox based. I'll look at this on Monday!

aredridel:
Mmm. I think this actually is eliminated by #1407 where I use flexboxes a bit better. Want to take a quick peek and see the site currently reflects something like your design sense?

vramana:
@aredridel The logout and profile aren't exactly aligned in the with profile photo. Also I think they are bit too close to each other which makes it easy to click the other one by mistake. Probably its better to arrange them horizontally. This will also make the npm logo top align with the profile-photo top. Currently profile-photo top is above the npm-logo.

PR #1401, opened 2015-10-17
fix includeSubdomains typo

tetsuo:
Looks like this should be `includeSubDomains` (http://hapijs.com/api)

### DIFF
--- a/lib/startup.js
+++ b/lib/startup.js
@@ -24,7 +24,7 @@ function makeServer(config) {
           security: {
             hsts: {
               maxAge: 1000 * 60 * 60 * 24 * 30,
-              includeSubdomains: true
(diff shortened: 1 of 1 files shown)


aredridel:
Sure enough! Thanks for catching this! Looks like it was documented wrong in hapi at one point, since fixed!

aredridel:
... Whaaaat.  That actually crashes! Despite being documented!

aredridel:
... Because that's in hapi 11, where this spelling is correct for 9. Augh!

tetsuo:
Ah, my mistake, should've checked the version.

Looks like this is fixed very recently (https://github.com/hapijs/hapi/pull/2776)

PR #1402, opened 2015-10-17
let humans read time values

tetsuo:
Parse pretty time strings with https://github.com/nescalante/timeparse so humans can also read them


aredridel:
Interesting! That makes me a bit uncomfortable, introducing a new dependency and a new language, but it's a pretty controlled use. 

How's this feel to you, @jefflembeck, @jonathanmarvens?

tetsuo:
This also can be done with `moment` indeed. See: http://momentjs.com/docs/#/durations/

jefflembeck:
It adds a dependency for a convenience method that looks like it'd get used only a couple of times, one of which is already supported by a very readable variable.

Issue #1403, opened 2015-10-17
Issues with node 4.0/engine specifications needed


tekd:
There is a possible bug for node 4.0--when I run it the app crashes. It runs with node 0.12.7. Should engines be specified in the package.json file?

aredridel:
Yes, right now the site requires 0.10 or. 0.12. The oniguruma library needs porting. A PR would be accepted to add an engines spec.

taraadiseshan:
I'm working on a pull request!

ashleygwilliams:
:tada:

PR #1406, opened 2015-10-18
added engine

taraadiseshan:
addresses #1403

but it looks like engines will only throw a warning upon install.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,9 @@
       "hbsfy"
     ]
   },
+  "engines" : { 
+    "node" : ">=0.10 <4" 
+  },
   "dependencies": {
     "@soldair/marky-markdown": "^6.0.0",
     "accept-language-parser": "^1.0.2",


taraadiseshan:
Oh oops, I didn't see this before making the PR. Going back to it now, thanks

aredridel:
Yes, engines only warn. I'm working on mad science sorts of tricks to let npm install node itself, but haven't found satisfying options quite yet.

aredridel:
(there's an option you can add to your npmrc to make engine failure fatal)

aulvi:
Engines appears to only warn during package install.

So, if I clone a repo and it has a package.json with engines defined, running `npm install` or `npm start` doesn't return any engine warnings.

aredridel:
Oh that is indeed a little distressing, @aulvi! And that might be a change in npm@3.

aulvi:
@aredridel Could you provide an example of how to enforce the engine using a project-local .npmrc?

aredridel:
@aulvi `engine-strict=true`

I urge you not to do this though since it's a case of 'author knows better than user, user can't even try to make it go'

ashleygwilliams:
hey! so i'm p sure what's happening here is this:

- engine set in package
- npm install
- EXPECTATION: warn on engine mismatch

instead what happens
- engine declared in package
- package set as dependency
- npm install
- throw error on engine mismatch

the key thing here is that the engine should warn on npm install for the package it is declared in, as well as when it is used as a dep

hopefully this clarified!

aulvi:
:+1:

othiym23:
Just so everyone knows, `engineStrict` was nerfed in `npm@3`. From [the release notes](https://github.com/npm/npm/releases/tag/v3.0.0):

> ##### `engineStrict`
> 
> * [#6931](https://github.com/npm/npm/issues/6931) The rarely-used `package.json` option `engineStrict` has been deprecated for several months, producing warnings when it was used. Starting with `npm@3`, the value of the field is ignored, and engine violations will only produce warnings. If you, as a user, want strict `engines` field enforcement, just run `npm config set engine-strict true`.
> 
> As with the peer dependencies change, this is about shifting control from module authors to application authors. It turns out `engineStrict` was very difficult to understand even harder to use correctly, and more often than not just made modules using it difficult to deploy.

othiym23:
...so, to complete the thought, putting `engine-strict` in a per-project `.npmrc` is only going to work for people who are cloning a repo and running `npm install` directly (because npm no longer bundles `.npmrc` in packages, for security reasons), which should be fine for the web site but not so much for things meant to be used as libraries or plugins.

aredridel:
But that said, I think this engines declaration in this PR looks great, and I'll test and merge this on Monday!

aredridel:
Thank you!

taraadiseshan:
thank you! I enjoyed learning more about this :)

PR #1407, opened 2015-10-19
Tighten up whitespace a bit in the top bar, and tighten up the implementation a lot

aredridel:
Bonus: Fixes some alignment issues on mobile size!


jefflembeck:
Oh that looks _so_ much better.

Issue #1399, opened 2015-10-16
Private modules signup flow is hidden for orgs-capable users


seldo:
If signed up with an orgs-capable but unpaid user, clicking "sign up for private modules" takes you to your "subscriptions" page which lists no subscriptions. There are no instructions on how to sign up for private modules.

If you go to "payment info" tab it says "private modules cost $7 per user" and entering payment info gives me a private modules account.

1. "Payment info" by itself should not be signing me up for a product, maybe shouldn't exist until there *is* payment info.
2. Clicking "sign up for private modules" should take me directly to a page that *does* sign me up for private modules

This is mostly about the copy on those pages and the destination of the links, not a major functional issue.

rockbot:
I believe this is being taken care of via the issues listed in https://github.com/npm/issue-tracker/issues/308

PR #1413, opened 2015-10-19
Language fixing + link fixing for private modules

jefflembeck:
This should address #1399 

I'd like a look-over on this from @seldo / anybody who should be in charge of language


jefflembeck:
Got the :+1: from Kasey on the change.

aredridel:
Looks right!

aredridel:
:+1:

PR #1414, opened 2015-10-20
Only show controls to those permitted

jefflembeck:
The form fields were visible for other org users.
They are unable of actually doing anything with those fields, but the UI
is confusing and misleading


aredridel:
LGTM!

PR #1416, opened 2015-10-20
New look for tab-based navigation

jefflembeck:
We've added nav-4, nav-4 is a tab-style nav that we will be using for
the org screen. It has been created, added to the ui-docs and to the org
page.

This commit also cleans up a bunch of the nav-related css


aredridel:
How does this work in the absence of javascript?

jefflembeck:
So, tabs are something we already have in place and are just deep-links that are collapsed currently. This is just another visual way of representing them.

PR #1417, opened 2015-10-21
Set up newrelic

jefflembeck:
Trying out newrelic to see what it can give us on the browser end.

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,4 +1,5 @@
 require("./lib/environment")();
+require("newrelic");
 
 var replify = require('replify');
 var bole = require('bole');
--- a/package.json
+++ b/package.json
@@ -25,8 +25,8 @@
       "hbsfy"
     ]
   },
-  "engines" : { 
-    "node" : ">=0.10 <4" 
+  "engines": {
+    "node": ">=0.10 <4"
   },
   "dependencies": {
     "@soldair/marky-markdown": "^6.0.0",
@@ -69,6 +69,7 @@
     "mousetrap": "0.0.1",
     "murmurhash": "0.0.2",
     "mustache-mailer": "^2.1.2",
+    "newrelic": "^1.22.2",
     "nib": "^1.1.0",
     "node-uuid": "^1.4.3",
     "node-zendesk": "^1.1.2",


server.js
```diff
@@ -1,4 +1,5 @@
 require("./lib/environment")();
+require("newrelic");
```
aredridel: Wait, this sounds like backend tracing, not frontend.

This loads continuation-local-storage and ... eek!

newrelic.js
```diff
@@ -0,0 +1,24 @@
...
+  license_key: process.env.NEW_RELIC_LICENSE_ID,
+  logging: {
+    /**
+     * Level at which to log. 'trace' is most useful to New Relic when diagnosing
+     * issues with the agent, 'info' and higher will impose the least overhead on
+     * production applications.
+     */
+    level: 'info'
+  }
+}
```
rockbot: is this file at the root? should it maybe go into /lib or /adapters?

aredridel:
Other than the conceptual horror at loading CLS, looks good to me.

server.js
othiym23: What's wrong with `continuation-local-storage`? :<

Also,

1. Most versions of `newrelic` don't use CLS or `async-listener`. The team rewrote the base instrumentation layer to remove several levels f indirection for performance reasons.
2. The dance to set up RUM is kind of complicated, and while it can be done manually, using the Node agent to configure and load the browser agent is way simpler, and will also give better info because it has the server-side context.

Given that the NR Node agent team keeps finding (& trying to fix) leaks in Node core, I'm pretty confident this won't be too much of a resource hog, but I can work with the team if y'all decide that you can't even with the Node monitoring. We'll be on our own for support at that point, though.

aredridel: Oh, I was not aware! Last time I used new relic, I did end up debugging CLS. I figured it was inherent!

aredridel: (mostly, CLS just introduces new things that you have to be aware of, globally. PayPal had a fair number of places where we'd lose the context.)

newrelic.js
othiym23: It needs to be in the app root for the agent to bootstrap, unless you want to set the app name and license info as environment variables, in which case you don't need this file at all. (You can also set the location of the config file with environment variables, but why would you do that.)

PR #1421, opened 2015-10-21
replace deprecated hgroup elements

### LINKED ISSUE #845: hgroup is deprecated
This is a very minor issue... I noticed when looking through static-pages that there were a few `hgroup`s in the code. This element was [deprecated](https://www.w3.org/wiki/HTML/Elements/hgroup) (see Ian Devlin's [post](http://www.iandevlin.com/blog/2013/04/html5/hgroup-removed-from-the-html5-specification)). This doesn't affect functionality at all.

Also filed under npm/static-pages#11

tyleranton:
This PR replaces the deprecated `hgroup` elements with `header` elements.

Fixes #845

### DIFF
--- a/assets/styles/reset.styl
+++ b/assets/styles/reset.styl
@@ -5,7 +5,7 @@ small, sub, sup, var, b, i, dl, dt, dd, ol, ul, li,
 fieldset, form, label, legend,
 table, caption, tbody, tfoot, thead, tr, th, td,
 article, aside, canvas, details, figcaption, figure,
-footer, header, hgroup, menu, nav, section, summary,
+footer, header, menu, nav, section, summary,
 time, mark, audio, video
   margin 0
   padding 0
@@ -15,7 +15,7 @@ time, mark, audio, video
   vertical-align baseline
 
 article, aside, details, figcaption, figure,
-footer, header, hgroup, menu, nav, section
+footer, header, menu, nav, section
   display block
 
 hr
--- a/assets/styles/_header.styl
+++ b/assets/styles/_header.styl
@@ -386,20 +386,6 @@ body>nav
     display inline-block
     max-width 140px
 
-  hgroup
-    display block
-    margin 0 0 30px 0
-    h2,h3
-      text-align center
-    h2
-      color white
-      font-size 36px
-      font-weight 600
-    h3
-      @extend .featherweight-heading
-      font-size 24px
-      color rgba(white, 0.8)
-
   a.button
     color rgba(background-color, 0.8)
     background-color darken(npmred, 10%)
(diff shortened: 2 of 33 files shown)


tyleranton:
Ah missed the tests in `package.js`.

Do they need to be modified or removed?

tyleranton:
With replacing `hgroup` with `header`, what should the two tests and the styles be modified to look for?

aredridel:
That would make sense -- both `hgroup` and `header` have little default style and are block elements. You may need to make sure you don't run into existing header styles, but they should be well isolated.

tyleranton:
Looks like there is one existing `header` element in `default.hbs` on `line 49`.

aredridel:
That's the one. It's the styles that might be aimed at that would land elsewhere that I worry about.

tyleranton:
Yea for sure. Any suggestions?

ashleygwilliams:
hey @tyleranton - i can help you through this tomorrow! i'm ag_dubs on twitter (open DMS), freenode IRC, and mozilla IRC

tyleranton:
Okay, thanks!

aredridel:
Just checking in -- did you poke this at all? I like this change :)

tyleranton:
Yea I'm not sure what to do about that one pre-existing header element. I also need guidance on what I should do to modify those two test cases.

ashleygwilliams:
h'ok @tyleranton here's my thought on a solution:

- instead of replace `hgroup` with header (that has it's own styles), replace with `div` with a class such as `titlegroup`
- instead of deleting `hgroup` styles, target those styles at the `titlegroup` class, e.g. ".titlegroup`
- in the tests, check for the class not the element type

i'll be on irc at 3:30 to answer any of your q's as promised, but i think this general strategy should lead you to success :) (no promises though as i haven't actually tried it myself)

tyleranton:
...*ahem*... YAAAAAAASSSSSS

(sorry, that felt necessary)

assets/styles/reset.styl
```diff
@@ -5,7 +5,7 @@ small, sub, sup, var, b, i, dl, dt, dd, ol, ul, li,
 fieldset, form, label, legend,
 table, caption, tbody, tfoot, thead, tr, th, td,
 article, aside, canvas, details, figcaption, figure,
-footer, header, hgroup, menu, nav, section, summary,
```
ashleygwilliams: i would add `.titlegroup` here (just to be safe)

assets/styles/reset.styl
```diff
@@ -15,7 +15,7 @@ time, mark, audio, video
   vertical-align baseline
 
 article, aside, details, figcaption, figure,
-footer, header, hgroup, menu, nav, section
```
ashleygwilliams: here too

PR #1427, opened 2015-10-22
Add support for an Org's "full name"

jefflembeck:
fullname is an optional attribute for an organization that is currently
stored in the user-acl `resource` hstore (as `human_name`) but will eventually become a
column on that model.

Since it is stored as `human_name` on the user-acl, we finesse it a
little along the way.

This updates tests, fixtures, handlers, and the agent

### DIFF
--- a/agents/org.js
+++ b/agents/org.js
@@ -21,8 +21,8 @@ Org.prototype.create = function(opts, callback) {
   var url = USER_HOST + '/org';
   var resource = {};
 
-  if (opts.fullname) {
-    resource.human_name = opts.fullname;
+  if (opts.human_name) {
+    resource.human_name = opts.human_name;
   }
 
   var data = {
--- a/agents/org.js
+++ b/agents/org.js
@@ -15,19 +15,21 @@ var Org = module.exports = function(bearer) {
   this.bearer = bearer;
 };
 
-Org.prototype.create = function(scope, fullname, callback) {
+Org.prototype.create = function(opts, callback) {
+  opts = opts || {};
+
   var url = USER_HOST + '/org';
   var resource = {};
 
-  if (fullname) {
-    resource.human_name = fullname;
+  if (opts.fullname) {
+    resource.human_name = opts.fullname;
   }
 
-  var opts = {
+  var data = {
     url: url,
     json: true,
     body: {
-      name: scope,
+      name: opts.scope,
       resource: resource
     },
     headers: {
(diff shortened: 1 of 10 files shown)


agents/org.js
```diff
@@ -15,14 +15,20 @@ var Org = module.exports = function(bearer) {
   this.bearer = bearer;
 };
 
-Org.prototype.create = function(name, callback) {
+Org.prototype.create = function(scope, fullname, callback) {
```
aredridel: It might be nice to either now or eventually make these parameters an object.

jefflembeck: :+1: Good call!

aredridel:
This looks great.

handlers/customer.js
```diff
@@ -288,32 +289,34 @@ customer.subscribe = function(request, reply) {
               })
           } else {
             return Org(loggedInUser)
-              .create(planInfo.npm_org);
+              .create({
+                scope: planInfo.npm_org,
+                fullname: planData.fullname
```
rockbot: on line 228, we've got `planInfo.npm_org = planData.orgScope;`, and then on 293 we've got `scope: planInfo.npm_org,`

maybe tighten it up by simply making `scope: planData.orgScope`? 

this then also opens up the opportunity to simply create with the planData object :-)

handlers/customer.js
```diff
@@ -288,32 +289,34 @@ customer.subscribe = function(request, reply) {
...
-                    }
-                    return accept();
-                  });
+        }).then(function() {
+          return Customer(loggedInUser).createSubscription(planInfo)
+            .tap(function(subscription) {
+              if (typeof subscription === 'string') {
+                request.logger.info("created subscription: ", planInfo);
+              }
+              return new P(function(accept, reject) {
```
rockbot: I really prefer accept/reject over resolve/reject... note to self: turn all resolve/rejects into accept/reject :-)

handlers/customer.js
jefflembeck: :+1: oh yes. coming right up.

handlers/org.js
```diff
@@ -33,6 +33,20 @@ exports.getOrg = function(request, reply) {
           }
         }
 
+        org = org || {};
+        org.info = org.info || {};
+        var fullname = "";
+
+        if (org.info.human_name) {
+          fullname = org.info.human_name;
```
rockbot: is this in preparation for some future time when human_name is at the top level of org information (i.e. not a resource)?

aredridel: Hmm. `info` seems as arbitrary as `resource` so I'd tend to just expose `resource` until it's refactored.

jefflembeck: @rockbot yes. It's a little pre-emptive. I can drop it.

jefflembeck: @aredridel `.info` is already on the `org`, coming in from the agent, so I put it on there, so as to avoid future refactorings I'm going to have to do.

jefflembeck: that being said, since they're going with `human_name` on the user-acl, I can just rip out all `fullname` references in our stuff.

aredridel: Shiny!

jefflembeck:
Part of this is https://github.com/npm/newww-tests/pull/14

jefflembeck:
Coming back to this to fix a few things, namely `human_name` doesn't follow our styles at all, let's curb that tomorrow morning.

PR #1430, opened 2015-10-24
Change "node" to "Node.js" ...

jonathanmarvens:
We mention __Node.js__ as "__node__" many times in the codebase and we're a bit inconsistent with it, so I chased the instances down. Obviously not a super important PR, but I just thought, "Why not?"

CC: @aredridel, @rockbot, @jefflembeck.

### DIFF
--- a/templates/homepage.hbs
+++ b/templates/homepage.hbs
@@ -1,7 +1,7 @@
 <div class="container">
 
   <div class="centered">
-    <h1>npm is the package manager for <span id="what-npm-is-for">node</span></h1>
+    <h1>npm is the package manager for <span id="what-npm-is-for">node.js</span></h1>
   </div>
 
   <div id="home-stats">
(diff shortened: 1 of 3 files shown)


templates/homepage.hbs
```diff
@@ -1,7 +1,7 @@
 <div class="container">
 
   <div class="centered">
-    <h1>npm is the package manager for <span id="what-npm-is-for">node</span></h1>
+    <h1>npm is the package manager for <span id="what-npm-is-for">node.js</span></h1>
```
jonathanmarvens: Left this in lowercase, because I noticed everything here is in lowercase.

aredridel: Yeah. Not sure if node has a canonical capitalization but npm sure is lowercase.

aredridel:
Thoughts? @ashleygwilliams 

this starts getting into how we want to refer to node.

ashleygwilliams:
hayy-- Node.js calls themselves Node.js. i think Node.js gets to be a bit much, but i think keeping Node capitalized is a good call (as much as i hate capitalization)

i like this standarization of language! gonna go through the docs and do this as well! (or might make it a "your-first-pr" issue)

templates/homepage.hbs
ashleygwilliams: I'd capitalize it.

jonathanmarvens:
@ashleygwilliams “*i think Node.js gets to be a bit much, but i think keeping Node capitalized is a good call (as much as i hate capitalization)*” - But is it really up to us to decide it’s okay to strip out the “__.js__”. I know this may seem picky, but I always choose accuracy and consistency in situations like this … you see where I’m coming from?

ashleygwilliams:
yup @jonathanmarvens. i guess i meant to say i think we should go with `Node.js` unless ya'll think i'm being a stinker in which case `Node`... but it seems we agree more that i assumed! let's go with `Node.js`

aredridel:
LGTM!

jonathanmarvens:
Understood, @ashleygwilliams! No worries.

PR #1434, opened 2015-10-26
adds ability for logged-in users to manage their cli tokens from the website

rockbot:
This is a proof of concept, with the most basic of styling. I'm sure @ncawthon will have Opinions as to what this should look like long-term ;-)


routes/authenticated.js
```diff
@@ -48,6 +48,14 @@ module.exports = [
     method: "GET",
     handler: require('../facets/user/show-email-edit')
   }, {
+    path: "/security",
+    method: "GET",
+    handler: require('../handlers/user').getCliTokens
+  }, {
+    path: "/token/{token}",
+    method: "POST",
```
aredridel: We may want to handle DELETE here for completeness.

aredridel:
This looks fantastic and well structured.

routes/authenticated.js
rockbot: i'd like @jefflembeck's thoughts on this as well - I matched the pattern he used for adding/removing users from an org.

in particular, can forms handle DELETEs across all browsers?

jefflembeck: forms cannot handle DELETEs across all browsers - Unless we're talking adding DELETE here for the sake of a `curl` tool, which then references the same thing?

seldo:
\o/ for this.

routes/authenticated.js
aredridel: Yeah, or future AJAXy things. I'd add delete, not use just delete.

rockbot: no, this endpoint is hit via a form; as such, i think we need to stick with a POST :-(

jefflembeck:
:sheep:

routes/authenticated.js
aredridel: Yeah. We can do both! POST as the "no real semantics to this request, just do a thing". But I'm happy shipping it as is too

aredridel:
:sheep:

aredridel:
LGTM!

Also yay getting rid of `e. g.`

PR #1435, opened 2015-10-26
Make search box autofocus

aredridel:


aredridel:
Well fudge. This irrecoverably destroys accessibility via text to speech.

PR #1440, opened 2015-10-27
Fix nesting for org

jefflembeck:
Incorrect handlebars nesting was making it so deleting a user wasn't working.


aredridel:
LGTM!

PR #1441, opened 2015-10-27
Handle 500 delete

jefflembeck:
We've got a 500 on production and it needs handling.

This change involved some serious promisifying of our handler, so I could better track the errors - 4xx errors are now displayed as notices while 5xx errors are sent to an internal page.

Test added for 500.

### DIFF
--- a/agents/org.js
+++ b/agents/org.js
@@ -327,11 +327,10 @@ Org.prototype.removeUser = function(name, userId, callback) {
       }
 
       if (resp.statusCode >= 400) {
-        err = new Error(body);
+        err = new Error(removedUser);
         err.statusCode = resp.statusCode;
         return reject(err);
       }
-
 
       return accept(removedUser);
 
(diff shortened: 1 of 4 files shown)


agents/org.js
```diff
@@ -306,32 +306,35 @@ Org.prototype.getTeams = function(name, callback) {
...
+      if (resp.statusCode >= 400) {
+        err = new Error(removedUser);
+        err.statusCode = resp.statusCode;
+        return reject(err);
+      }
 
+      return accept(removedUser);
+
+    });
+  }.bind(this)).nodeify(callback);
```
rockbot: the bind pattern isn't something we use elsewhere in the codebase (we generally use `var self = this;`)... I don't mind it, but think we should be consistent... thoughts?

aredridel: I'd normally just pass the property I want: `var bearer = this.bearer` and later, just use the free `bearer` variable.

aredridel:
LGTM as well.

jonathanmarvens:
LGTM too, @jefflembeck.

PR #1448, opened 2015-10-29
CMS Integration

aredridel:
Depends on and includes #1450


agents/cms.js
```diff
@@ -0,0 +1,139 @@
...
+    });
+  } else {
+    debug("Request for %j is already pending", key);
+  }
+
+  return cache.pending[key];
+};
+
+function withRedis(liftedFn) {
+  var redisP = pool.acquireAsync();
```
chrisdickinson: Hm, my only concern is if hapi uses domains, some manual work might be needed to re-associate the redis client with the right domain. There's some prior art in [pg](https://github.com/brianc/node-postgres/blob/master/lib/pool.js#L75-L89), which AFAICT does this correctly.

aredridel: Oh dear. Let me investigate. If it's wrong, what happens?

templates/layouts/default.hbs
```diff
@@ -207,6 +211,9 @@
   </footer>
   {{#if devEnv}}<script defer async src="/static/js/tota11y.min.js"></script>{{/if}}
   <script defer src="/static/{{m "js/vendor.min.js"}}"></script>
+  {{#each page.scripts}}
+  <script defer async src='{{this}}'></script>
+  {{/each}}
```
rockbot: do you need to update the CSP to accept these new scripts?

aredridel: Oh dear the CSP. OMG.

agents/cms.js
chrisdickinson: The failure mode can vary — I'm not sure what it would be like in hapi, though it might be pretty innocuous. I've seen similar problems create memory leaks in other frameworks, at least.

aredridel: So I don't think there's a problem here -- every retained callback is one set up early, outside of request context. cached promises have a finite lifetime, and this shouldn't leak.

templates/layouts/default.hbs
aredridel: Added 'https://assets.npmjs.com/' to CSP

aredridel:
I think this is done!

PR #1449, opened 2015-10-30
redirects orgs pages to /org/{org} from /orgs/{org}

rockbot:
what it says :-)


aredridel:
So. Since there's that nasty header-splitting bug in node core's http, we can't pass user input to redirect -- we have to make sure that the org name could possibly be valid before redirecting. If not, we should 400 or 404.

aredridel:
Shiny! Sheep it!

PR #1452, opened 2015-10-30
Only make a metrics emitter if there is config asking for one.

ceejbot:
If process.env.METRICS_URL is set, we make a real emitter. Otherwise, we make a stub that does nothing when called. This allows us to change the configuration in production by changing settings in etcd, without having to change code. It also allows us to not try to connect to a non-existent metrics collector in dev or test.

To get metrics flowing in production again, we'll need to accompany this change with

`renv config set -e production -a newww service.env.METRICS_URL=whatever`


aredridel:
Excellent. :sheep:

PR #1458, opened 2015-11-02
Promisify all the things.

jonathanmarvens:
I'm working on promisifying more of the codebase. This PR will be the home of some of those commits.

### DIFF
--- a/lib/startup.js
+++ b/lib/startup.js
@@ -59,17 +59,15 @@ function makeServer(config) {
         cookie: process.env.SESSION_COOKIE,
         clearInvalid: true,
         validateFunc: function(session, cb) {
-          cache.get(session.sid, function(err, item, cached) {
-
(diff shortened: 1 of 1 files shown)


lib/startup.js
```diff
@@ -59,17 +59,15 @@ function makeServer(config) {
...
-
-            if (err) {
-              return cb(err, false);
-            }
+          P.promisify(cache.get, cache)(session.sid).catch(function(err) {
+            cb(err, false)
+          }).spread(function(item, cached) {
             if (!cached) {
-              return cb(null, false);
+              cb(null, false)
```
aredridel: You probably want to learn about promise `.nodeify(cb)` -- which will pass error or result to the callback from a promise (and if cb is null, do nothing), letting you have a function that returns a promise _or_ takes a callback

PR #1459, opened 2015-11-02
Promisified the internals of the `validateFunc` option …

jonathanmarvens:


lib/startup.js
```diff
@@ -59,16 +59,14 @@ function makeServer(config) {
         cookie: process.env.SESSION_COOKIE,
         clearInvalid: true,
         validateFunc: function(session, cb) {
-          cache.get(session.sid, function(err, item, cached) {
-
-            if (err) {
-              return cb(err, false);
-            }
+          P.promisify(cache.get, cache)(session.sid).catch(function(err) {
```
aredridel: Hm. What if `cache.get` returned a promise to start? that'd make this a lot tidier.

aredridel: Unless of course we didn't write that.

jonathanmarvens: @aredridel Yes, it’s returned by [hapi.js](http://hapijs.coms.com "hapi.js")’s [`Server.prototype.cache(...)`](http://hapijs.com/api/9.3.1#servercacheoptions "9.3.1 API Reference").

aredridel:
Given the above, looks good to me!

PR #1460, opened 2015-11-03
Org permission breakdown

rockbot:
we need a way to only present information to people who are allowed to see it.

thoughts? opinions?


jefflembeck:
:sheep:

omg yes.

aredridel:
Oh yes!

PR #1461, opened 2015-11-04
uses Boom to give more info to http errors

rockbot:
this is only for one file, but I think it's a good way to get more understandable logs

### DIFF
--- a/models/user.js
+++ b/models/user.js
@@ -9,6 +9,7 @@ var P = require('bluebird');
 var Request = require('../lib/external-request');
 var userValidate = require('npm-user-validate');
 var utils = require('../lib/utils');
+var Boom = require('boom');
 
 var chimp;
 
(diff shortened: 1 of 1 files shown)


aredridel:
LGTM if body is a string

aredridel:
Yeah, that makes sense and is generally safe.

models/user.js
```diff
@@ -185,9 +185,8 @@ User.prototype.getPackages = function(name, page, callback) {
         return reject(err);
       }
       if (resp.statusCode > 399) {
-        var err = new Error('error getting packages for user ' + name);
-        err.statusCode = resp.statusCode;
-        return reject(err);
+        err = new Error('error getting packages for user ' + name);
+        return reject(Boom.wrap(err, resp.statusCode, JSON.stringify(body)));
```
aredridel: Hmm. I'm reading more closely now -- maybe we want `Boom.create` instead of making an error and immediately wrapping it. The create function takes three arguments -- `statusCode`, `message` and `data` -- perhaps the body can be the data, and we can supply a contextual message?

jefflembeck:
Any chance we'll want to try/catch the JSON.stringify? (I don't foresee circular references happening in this case, but...)

PR #1463, opened 2015-11-06
Ui docs

ncawthon:
Thanks @jonathanmarvens with your help with the variables and the debugging.

Need to still scrub the .styl refs to remove any explicit red / blue / whatevs class names. This will come in handy if we ever co-brand from a palette perspective.


aredridel:
LGTM!

PR #1465, opened 2015-11-09
Node.js 4 upgrade.

jonathanmarvens:


jonathanmarvens:
Okay to merge? CC: @aredridel, @jefflembeck, @rockbot.

aredridel:
Rollout plan is gonna take some care, actually, because we have node 0.10 in production.

aredridel:
So we should talk about how to do this ASAP.

aredridel:
Ooh. Actually, can you push 5f559b6 to staging?  www-staging-1.internal.npmjs.com has node 4.0.0 (should bump that, but 4.0.0 should work, I think) and we can see if it works there.

aredridel:
Something else to consider is delaying the hapi 11 update until after node 4 -- hapi 9 will warn, but I believe it won't fail on node 0.10. That way we can get a node 4 compatible site out, test against both, and then update production to 4, then update to hapi 11, since that's not a reversible change we can run on 0.10

aredridel:
Come on Travis!

aredridel:
LGTM! :tada:

jonathanmarvens:
Awesome :smile:!

adapters/metrics.js
```diff
@@ -13,11 +12,9 @@ module.exports = function constructEmitter(app) {
   else {
     emitter = new Emitter({
       uri: process.env.METRICS_URL,
-      node: os.hostname(),
-      app:app, // optional.
+      app: app,
+      port: process.env.PORT || '0'
```
aredridel: I'm curious about this bit -- what's the setting for?

jonathanmarvens: @aredridel This was @ceejbot's edit, after updating __npm/newww__ to use [__numbat-emitter v2.0.3__](https://github.com/ceejbot/numbat-emitter/tree/v2.0.3 "ceejbot/numbat-emitter at v2.0.3"). She said the API changed a bit [from __v1.1.3__ to __v2.0.3__](https://github.com/ceejbot/numbat-emitter/compare/v1.1.3...v2.0.3 "Comparing v1.1.3...v2.0.3 · ceejbot/numbat-emitter"). Maybe she can comment a bit on this when she's back?

ceejbot: See the numbat-emitter docs: https://github.com/ceejbot/numbat-emitter#configuration 

If you look at registry service metrics you can see examples of host & port fields being used in the graphs to distinguish processes on a single box. actually this needs one more change: you will want to make sure that `app` is set to something because www doesn't otherwise namespace its metrics. It's junking up the top-level namespace in the staging influxdb as a result. Visit http://influxdb-1-west.internal.npmjs.com:8083/?# & run `SHOW MEASUREMENTS` on the `registrystaging` db to see.

`app: app || 'www',` on that line would do it.

aredridel: Oh, excellent, thank you, that explains it. And then yes, let's get that change in.

aredridel:
Whoops! Other merges stepped on this one's toes. Time for a rebase or fixup?

jonathanmarvens:
Woohoo!!!

PR #1467, opened 2015-11-09
Add assets.npmjs.com more broadly to the CSP

aredridel:


aredridel:
Removed the silly specific test.

PR #1469, opened 2015-11-10
Team page packages

rockbot:
those team page packages. so hot right now.


jefflembeck:
Instead of having two paths, since we get the package and member data in the same call, we should probably treat the team page as one page with two tabs, and then use our tab component. I suppose this could be for the next PR

jefflembeck:
Perfect! :sheep:

PR #1472, opened 2015-11-10
Updated the "async" dependency.

jonathanmarvens:


jonathanmarvens:
Tested and nothing breaks. It seems we're only using `async.parallel` anyways, which shouldn't have had any breaking changes.

aredridel:
LGTM :)

PR #1473, opened 2015-11-10
Updated the "stylus" dependency.

jonathanmarvens:


aredridel:
LGTM and builds styles properly.

PR #1474, opened 2015-11-10
Updated the "cheerio" dependency.

jonathanmarvens:


jonathanmarvens:
AFAICT, `cheerio` is only used in the tests and all of the tests pass, so I think we're good here.

aredridel:
Yup!

aredridel:
LGTM

PR #1485, opened 2015-11-11
Add user to team

jefflembeck:
This allows a user to add a member, or a list of members, in an org to a team.

It strongly lacks styles that are even remotely ok.

/cc @rockbot


agents/team.js
```diff
@@ -293,7 +293,7 @@ Team.prototype._addUser = function(opts, callback) {
       }
 
       if (resp.statusCode >= 400) {
-        err = new Error(body);
+        err = new Error(body.message);
         err.statusCode = resp.statusCode;
```
rockbot: is it `body.message`? or `body.error`? (this may be a good question for @chrisdickinson; i think errors coming from user-acl now have the form `{ error: "message" }` in the body...?

assets/scripts/add-users-to-teams.js
```diff
@@ -0,0 +1,190 @@
...
+    "</a></span>" +
+    "<span class='role'>" + role + "</span>" +
+    "<button type='button' class='delete-user icon-x'>" +
+    "<span class='a11y-only'>Remove User " +
+    username +
+    "</span></button>" +
+    "<input type='hidden' name='member' value='" +
+    username + "' />";
+
+};
```
rockbot: not totally necessary for this PR, but it might be interesting for you to look at https://github.com/npm/newww/blob/master/assets/templates/profile-package.hbs and https://github.com/npm/newww/blob/master/assets/scripts/fetch-packages.js, which will allow you to use hbs for these templating bits :-)

assets/scripts/add-users-to-teams.js
```diff
@@ -0,0 +1,190 @@
...
+        return;
+      }
+
+      return getTeamUsers(auf.orgScope, teamName)
+        .done(function(users) {
+          auf.removeAll();
+          return auf.addUsers(users);
+        })
+        .fail(function(response) {
+          console.log(response);
```
rockbot: do we want to send these errors to the console? or can we show them to the user?

handlers/org.js
```diff
@@ -561,4 +561,69 @@ exports.redirectToOrg = function redirectToOrg(request, reply) {
...
+  var orgName = request.params.org;
+
+  var username = request.query.member;
+
+  if (invalidUserName(orgName)) {
+    return reply({
+      error: "Org not found"
+    })
+      .code(404)
+      .type('application/json');
```
rockbot: ooooh - i didn't know about `.type()`! Neat :-D

handlers/team.js
```diff
@@ -215,9 +230,118 @@ exports.updateTeam = function(request, reply) {
...
+      opts.team = team;
+      return reply.view('team/add-user', opts);
+    })
+    .catch(function(err) {
+      request.logger.error(err);
+
+      if (err.statusCode === 404) {
+        return reply.view('errors/not-found', err).code(404);
+      } else if (err.statusCode < 500) {
+        return handleUserError(request, reply, '/org/' + orgName, err.message);
```
rockbot: do you want to send users back to the org page? or to the team page?

templates/team/add-user.hbs
```diff
@@ -0,0 +1,51 @@
...
+            </div>
+            <ul class="members-list">
+            </ul>
+          </div>
+        </section>
+
+        <input type="hidden" name="updateType" value="addUsersToTeam"/>
+        {{> form_security}}
+        <button class="button" type="submit">Make it so</button>
+        <a href="/org/{{org}}">cancel</a>
```
rockbot: it might be good to redirect to the org page or the team page, depending on how the user was referred to the page (i.e. via the team creation process or via the "add members" button on the team page.

assets/scripts/add-users-to-teams.js
jefflembeck: Ooo.. I wanna do that during refactor time.

assets/scripts/add-users-to-teams.js
jefflembeck: Definitely a mistake.

handlers/team.js
jefflembeck: I think any non 404, non 500+ error that we'd get at this point has to do with not having access to the team, or something similar, so I figured I'd take them back to the org level.

templates/team/add-user.hbs
jefflembeck: Do we have a preferred way to handle `referer` in this case?

handlers/team.js
rockbot: ooh, interesting, and a fair point. i'm thinking from the perspective of starting on the team page, and clicking on the "add member" button... but i think you're right. a 404 for a person would actually be a 400, with a "you must first add the user to the org" error. I think this is good to go then :-)

jefflembeck:
OK, all things addressed!

PR #1486, opened 2015-11-12
Team page settings

rockbot:
though not _technically_ part of the requirements list, it was easy enough to add :-)


jefflembeck:
A++ work @rockbot, as always.

:sheep:

PR #1506, opened 2015-11-18
Add 1x1 tracking pixel for enterprise ULA agreement

aredridel:


jonathanmarvens:
@aredridel: Where does `trackSignedULA(…)` get called?

aredridel:
Oh whoops. that is actually from my previous, more complicated approach

jonathanmarvens:
Ah okay. LGTM now!

PR #1509, opened 2015-11-18
Team add packages

rockbot:
team and super admins can now add packages to their teams.

note: they can add either packages from other teams or any package that they have write access to (currently called "personal" packages). any/all help with naming would be greatly appreciated!

### DIFF
--- a/assets/scripts/add-packages-to-teams.js
+++ b/assets/scripts/add-packages-to-teams.js
@@ -58,7 +58,8 @@ var getPackage = function(orgName, pkg) {
 
 var AddPackageForm = function(form) {
   this.$el = form;
-  this.selectMenu = this.$el.find("[name=teams]");
+  this.teamSelect = this.$el.find("[name=teams]");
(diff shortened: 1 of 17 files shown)


assets/scripts/add-packages-to-teams.js
```diff
@@ -0,0 +1,204 @@
...
+  var html = "<div class='notice'>" + notification + "</div>";
+  $(".container.content").prepend(html);
+};
+
+module.exports = function() {
+  $(function() {
+    var form = $("[data-form-function=add-package-to-team]");
+
+    if (form.length) {
+      var auf = new AddPackageForm(form);
```
aredridel: Seems like a lot of this setup code belongs in the AddPackageForm class. But should work fine as is

aredridel:
Wow. Really, looks good. I've got thoughts about clean-up and such, but should be done as cross-cutting changes not specific to adding this, so let's :sheep: it

PR #1529, opened 2015-11-18
Get rid of 70% first column on org tables

aredridel:


aredridel:
Dug in the history and it was just "General orgs styling" as a commit.

PR #1532, opened 2015-11-18
Permission-ed buttons

rockbot:
not everyone gets to click on the fancy buttons. now you only see them if you have permission to :-)


seldo:
:ship:

PR #1533, opened 2015-11-18
from left to center, politically and ideologically

ncawthon:


seldo:
If the last-child rule is the same as all the other rows why not just delete the last-child rule entirely?

PR #1535, opened 2015-11-18
fixes org redirect on user add error (fixes #1514)

### LINKED ISSUE #1514: Adding a nonexistent user to an org redirect to wrong page after error
Repro:
1. I'm on https://staging.npmjs.com/org/seldo-orgs-1/members
2. I try to add "fdasfasdfadsfads" to the org
3. Hit Enter
Result: redirected to https://staging.npmjs.com/org/seldo-orgs-1 with an error message that the user doesn't exist.

Expected: same error, but redirected to `/members/`

rockbot:
also clarifies the error message and fixes tests so that we don't accidentally err in the same way in the future.

### DIFF
--- a/agents/org.js
+++ b/agents/org.js
@@ -235,7 +235,7 @@ Org.prototype.addUser = function(name, user, callback) {
       }
 
       if (resp.statusCode === 404) {
-        err = Error('org or user not found');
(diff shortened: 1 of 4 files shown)


seldo:
:ship:

PR #1541, opened 2015-11-18
1516 fix - not so weird lookin

ncawthon:
i think we're going to want to remove this Save someday (soon)


ncawthon:
seeing that this is happening on the add packages as well, not just the index page.

ncawthon:
@seldo this is now PR https://github.com/npm/newww/pull/1545

PR #1543, opened 2015-11-18
use mulitpackage format when adding pkgs to teams

chrisdickinson:
This prevents a couch dogpile from the user-acl side.

### DIFF
--- a/agents/team.js
+++ b/agents/team.js
@@ -199,19 +199,33 @@ Team.prototype.addPackage = function(opts) {
 
 Team.prototype.addPackages = function(opts) {
   opts = opts || {};
-  opts.packages = opts.packages || [];
-  var self = this;
-
-  var requests = opts.packages.map(function(pkg) {
-    return self.addPackage({
-      package: pkg.name,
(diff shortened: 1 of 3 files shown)


agents/team.js
```diff
@@ -199,19 +199,33 @@ Team.prototype.addPackage = function(opts) {
...
+    headers: {
+      bearer: this.bearer
+    }
+  }).spread(function (resp, body) {
+    if (resp.statusCode >= 400) {
+      throw _.extend(new Error(
+        resp.statusCode === 400 ? body.error :
+        resp.statusCode === 401 ? 'user is unauthorized to perform this action' :
+        resp.statusCode === 404 ? 'Team or Org not found' :
+        body.error || err.message
```
rockbot: `err` isn't defined anywhere - this will 500 if we ever get to this point :-)

(also, we haven't been promisifying `Request`, but it's not a bad way to go! :-))

chrisdickinson: Good catch! Fixing now.

PR #1544, opened 2015-11-18
Don't show link to add packages unless you can.

seldo:


templates/org/teams.hbs
```diff
@@ -61,7 +61,11 @@
               {{#if packages.count}}
               <a href="/org/{{../../org.info.name}}/team/{{name}}#packages">{{packages.count}}</a>
               {{else}}
-              <a href="/org/{{../../org.info.name}}/team/{{name}}/add-package">+ add packages</a>
+                {{#if perms.isAtLeastTeamAdmin}}
+                  <a href="/org/{{../../org.info.name}}/team/{{name}}/add-package">+ add packages</a>
+                {{else}}
+                  None
```
rockbot: do we want it to actually say "None" here?

seldo: I am open to other wording but it can't say "add packages" :-)

jonathanmarvens:
Agreed, @rockbot.

PR #1545, opened 2015-11-19
1516 weird lookin

ncawthon:
addresses both modal and table


seldo:
Verified that this does indeed look less weird.

ncawthon:
kk - adding hoverstate

seldo:
Marking further tweaks as post-GA.

PR #1549, opened 2015-11-19
adds link to add user to org when adding non-member to team 

### LINKED ISSUE #1513: Attempting to add a user not in the org to a team should include a link to do so
On team-creation page like https://staging.npmjs.com/org/seldo-orgs-1/team/contractors/add-user if I attempt to add a user not in an org, I get a flash message saying so, which is great! But the flash message should link to the page where I add people to the org, not force me to figure that out.

rockbot:
fixes #1513

.. this isn't my favorite way to fix this... we'll improve with client-side handlebars templates


seldo:
:sheep:

PR #1558, opened 2015-11-19
adds roles of users to org member page (fixes #1556)

### LINKED ISSUE #1556: display member roles on Org collaborators page
Orgs users have requested that member roles appear on the collaborators page (under the "members" tab). This sounds reasonable; currently you can only see member roles if you click into specific teams, even though you set the role on the collaborators page when adding new members. cc @rockbot

rockbot:
also removes the `date added` column (which was filled with just "date" because we don't actually have that data).


seldo:
:sheep:

PR #1569, opened 2015-11-23
1552 list flex

ncawthon:
sorry for the chaos commit... TL;DR here goes:
- normalizing tables across users / members / teams / orgs / packages
  - creation of a tables.styl file
- removed all Org marketing page stuffs - that's handled in WP
- breadcrumbin' the header in the Add pages
- general typographic style unification
- messing with that switch thing (a beast)
- adding an images.styl, which will normalize the avatar treatments
- the JS tweaks are minor, just format of which the lists are returned in HTML
- changed the subscription / token tables to reflect as well
  - still some cleanup needed there on the tabs


assets/templates/team-package.hbs
```diff
@@ -1,6 +1,6 @@
 <li data-name="{{pkgname}}">
   <span class='name'>
-    <a href='/package/{{pkgname}}'>{{pkgname}}</a>
+    <a href='/package/{{pkgname}}' target="_blank">{{pkgname}}</a>
```
rockbot: why the `target="_blank"` here?

templates/team/add-package.hbs
```diff
@@ -36,18 +36,19 @@
               <span class="add-number">0</span>
               <span class="unit">Packages</span>
             </div>
-            <button id="remove-all-added" class="remove-all-pkgs btn-link" type="button">remove all</button>
+            <a id="remove-all-added" class="remove-all-pkgs btn-link" type="button">remove all</a>
```
rockbot: why the switch from `a` to `button` here?

ncawthon: vertical alignment stuffs, most likely

assets/templates/team-package.hbs
ncawthon: I am hesitant to link out of this Add flow... If I've manicured a list of packages I want, and click off to a package detail page... poof! there goes my efforts. As a workaround to tooltip / expansion state, pop them out to a new window to allow them to ensure it is indeed the package they're looking for

rockbot: :+1: works for me!

PR #1575, opened 2015-11-25
disables ability to remove super admin from an org (fixes #1564)

### LINKED ISSUE #1564: restrict deletion of Super Admin from org
At this time, if the super admin deletes themselves from their org, they end up in a limbo state where they can't view the expected org tabs, and cannot be added back as super admin. 

I tried it myself, and here's how it plays out: I removed myself as super admin.  The org is still listed in my subscriptions, and payment is still happening (confirmed by the "manage billing" page) and this seems to be allowing the other members to use the org. It seems technically I still own the org, but can’t view any of the org tabs + can't add new members. When I go to my profile page, ```0 organizations``` is what shows up. 

Since there can only be 1 super admin, and we can't change it currently, it makes sense to disallow the deletion of this role. cc @chrisdickinson @rockbot

rockbot:
basically makes this happen: 

screen shot 2015-11-24 at 4 19 04 pm

it's not much, and until a fix is made in the user-acl, a super admin can still delete themselves via the cli.

thoughts? cc @seldo, @chrisdickinson, @ncawthon

### DIFF
--- a/handlers/org.js
+++ b/handlers/org.js
@@ -50,11 +50,12 @@ exports.getOrg = function(request, reply) {
         return user;
       });
 
-
       var isSuperAdmin = org.users.items.filter(function(user) {
         return user.role && user.role.match(/super-admin/);
       }).some(function(admin) {
-        return admin.name === loggedInUser;
+        var isSA = admin.name === loggedInUser;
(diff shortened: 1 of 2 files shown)


seldo:
:sheep:

PR #1580, opened 2015-11-26
Teams can now be deleted! Fixes #1574.

### LINKED ISSUE #1574: unable to remove teams
New teams show up as an artifact of converting private modules account to an organization account which is expected. However, users are unable to edit or remove those teams. 

This is likely caused by a `/` escaping issue. 

This [screencast](http://g.recordit.co/sDFqv0vUWE.gif) demonstrates the issue and shows the redirect to a 404.

jonathanmarvens:
Not mergeable just yet. Adding tests and error handling.

### DIFF
--- a/agents/team.js
+++ b/agents/team.js
@@ -416,7 +416,7 @@ Team.prototype.removeTeam = function(opts) {
       }
 
       if (resp.statusCode === 401) {
-        err = new Error('The user must be an admin to perform this operation.');
+        err = new Error('You do not have permission to perform this operation.');
         err.statusCode = resp.statusCode;
         return reject(err);
       }
--- a/agents/team.js
+++ b/agents/team.js
@@ -397,6 +397,32 @@ Team.prototype.removeUser = function(opts) {
   });
 };
 
+Team.prototype.removeTeam = function(opts) {
+  opts = opts || {};
+
+  var url = USER_HOST + '/team/' + opts.scope + '/' + encodeURIComponent(opts.id);
+
+  var data = {
+    body: {id: opts.id},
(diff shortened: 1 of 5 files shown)


jonathanmarvens:
Pushing to staging to make sure things are working the same way they are on my machine.

__CC:__ @rockbot, @ehsalazar.

agents/team.js
```diff
@@ -397,6 +397,47 @@ Team.prototype.removeUser = function(opts) {
...
+  };
+
+  return new P(function(accept, reject) {
+    Request.del(data, function(err, resp, body) {
+      if (err) {
+        return reject(err);
+      }
+
+      if (resp.statusCode === 401) {
+        err = new Error('The user must be an admin to perform this operation.');
```
aredridel: Is that the only case that gives a 401? A more generic 'you do not have permission' might be accurate longer

handlers/team.js
```diff
@@ -265,6 +265,7 @@ exports.updateTeam = function(request, reply) {
     var orgName = request.params.org;
     var teamName = request.params.teamName;
     var tab = '';
+    var section = '/team/' + teamName;
```
aredridel: This should work fine. However: I'd love for us as a team to consider using `url.resolve` whenever possible, rather than concatenating strings to form URLs. It knows the rules and should form more durable constructions.

aredridel:
Nice. Solid work! :sheep: it as is, though I'd questions about things that may become tech debt over time.

jonathanmarvens:
@aredridel: Yes, totally understood. We should discuss those things.

PR #1585, opened 2015-11-30
Fix broken package test

jefflembeck:
These changes came about because there was a failing test, but it wasn't failing when we ran the tests, despite a "failing" print out.

This happens when a test fails inside of a `catch` block.

[One](https://github.com/npm/newww/commit/796890374350a3ff5c6e4ddfdd77949c71962ad8): the formatter had a thing or two to say about our code at the time, so I made sure to run it before making other changes.

[Two](https://github.com/npm/newww/commit/1303e2e8ae432734e22179457dfeb9a42a6c72f8): Fixed the failing test

[Three](https://github.com/npm/newww/commit/2466fe307a2a9148844c9677ebff503f41e0cf87): Make sure this doesn't happen again. Now tests that fail in the `catch` blocks will stop the test runner

The problem with change three is that it's a global change. We should talk about where that goes.

### DIFF
--- a/test/agents/team.js
+++ b/test/agents/team.js
@@ -3,13 +3,36 @@ var Code = require('code'),
   lab = exports.lab = Lab.script(),
   describe = lab.experiment,
   it = lab.test,
+  afterEach = lab.afterEach,
   expect = Code.expect,
   nock = require('nock'),
   fixtures = require('../fixtures');
 
 var Team = require('../../agents/team');
 
+var rejectionMap = {};
+
+process.on('unhandledRejection', function(reason, promise) {
+  rejectionMap[promise] = reason;
+});
+
+process.on('rejectionHandled', function(promise) {
+  delete rejectionMap[promise]
+  ;
+});
+
 describe('Team', function() {
+  afterEach(function(done) {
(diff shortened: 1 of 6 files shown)


test/agents/team.js
```diff
@@ -388,7 +411,7 @@ describe('Team', function() {
         }).catch(function(err) {
           expect(err).to.exist();
           // i'd like this to show _all_ of the errors, not just the first one.
-          expect(err.message).to.equal('Team or Org or not found');
+          expect(err.message).to.equal('Team or Org not found');
           expect(err.statusCode).to.equal(404);
         }).then(function(packages) {
           expect(packages).to.not.exist();
```
chrisdickinson: It'd probably be good to `.return(null).next(done).catch(done)` (or the shorthand, `.return(null).then(done, done)` (vs. `.finally(() => done())`, so that the assertion error is propagated to the test runner and it can be reported against the proper test case.

aredridel:
@jefflembeck is this still relevant?

aredridel:
Closing in favor of a refocused https://github.com/npm/newww/pull/1840

Issue #1578, opened 2015-11-25
Extra warning for "Stop paying us" link if user is super-admin of an org


seldo:
If the user is the super-admin of any org, warn them before canceling that they will be locking everyone in that org out of the product.

aredridel:
@ncawthon this is in staging and production if you want to make further tweaks to the wording.

PR #1593, opened 2015-11-30
Warn clearly on stop all payments what this will do to orgs

### LINKED ISSUE #1563: missing date on billing page
on the billing part of the website, when you click “stop paying us”:

```
Your private subscription will remain active until
```

aredridel:
Fixes #1563, #1578

### DIFF
--- a/agents/customer.js
+++ b/agents/customer.js
@@ -56,29 +56,31 @@ Customer.prototype.getStripeData = function(callback) {
   var self = this;
   var stripeUrl = this.host + '/customer/' + self.name + '/stripe';
 
-  Request.get({
-    url: stripeUrl,
-    json: true
-  }, function(err, resp, stripeData) {
+  return new P(function(accept, reject) {
+    Request.get({
+      url: stripeUrl,
+      json: true
+    }, function(err, resp, stripeData) {
 
-    if (err) {
-      return callback(err);
-    }
+      if (err) {
+        return reject(err);
+      }
 
-    if (resp.statusCode === 404) {
-      err = Error('customer not found: ' + self.name);
-      err.statusCode = resp.statusCode;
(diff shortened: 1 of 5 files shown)


agents/customer.js
```diff
@@ -56,29 +56,31 @@ Customer.prototype.getStripeData = function(callback) {
...
+        err.statusCode = resp.statusCode;
+        return reject(err);
+      }
 
-    return callback(null, stripeData);
-  });
+      accept(stripeData);
+    });
+  }).nodeify(callback);
 };
```
rockbot: \o/ i love the gradual promisification of our code :-D

handlers/customer.js
```diff
@@ -8,6 +8,8 @@ var utils = require('../lib/utils');
 var validate = require('validate-npm-package-name');
 var invalidUserName = require('npm-user-validate').username;
 
+var xtend = require('xtend');
```
rockbot: how does `xtend` compare to `_.merge` or `hoek.merge`? (We use `hoek` in other parts of the codebase.)

aredridel: Very similar! Happy to use those, I just couldn't recall that _'s was called 'merge'. (and hoek. Oy. I can't even with that name.)

rockbot: totally fair ;-) I'm open to whatever; I just like consistency if/when possible :-)

apparently `_` also has `_.extend`, which isn't quite as deep?

aredridel: For sure. I'll get that after I fix this test.

agents/customer.js
aredridel: Yesss. It's getting so much better. I can't wait until we can rip out the nodeifys.

PR #1595, opened 2015-12-01
Show the link to sign up for private modules if they have no private modules subscriptions

aredridel:
May need more refined criteria -- @chrisdickinson? @ncawthon?


aredridel:
We should wait to ship this at least until we figure out the bad parts of how private modules and orgs play together.

aredridel:
Alright. I think this can be shipped :)

PR #1600, opened 2015-12-01
Pagination for Org packages

jefflembeck:
Much like the fix for #1596, this is temporary.

The method for grabbing packages is now grabbing the first 100, instead of the first 25 packages. This will hold us off until we refactor to better cover this.

### DIFF
--- a/test/agents/org.js
+++ b/test/agents/org.js
@@ -689,11 +689,13 @@ describe('Org', function() {
 
       Org('bob').getPackages('bigco')
         .catch(function(err) {
-          orgMock.done();
           expect(err).to.exist();
         })
         .then(function(packages) {
           expect(packages).to.not.exist();
+        })
+        .finally(function() {
+          orgMock.done();
           done();
         });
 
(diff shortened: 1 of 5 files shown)


test/agents/org.js
```diff
@@ -680,4 +680,52 @@ describe('Org', function() {
...
+    it('returns an error when one happens', function(done) {
+      var orgMock = nock('https://user-api-example.com')
+        .get('/org/bigco/package?per_page=100&page=0')
+        .reply(404);
+
+      Org('bob').getPackages('bigco')
+        .catch(function(err) {
+          orgMock.done();
+          expect(err).to.exist();
+        })
```
rockbot: tests like this are a bit odd, since it's not guaranteed that both the `.catch` and the `.then` are going to be fired. Instead, I like to use `.finally` to check that the mocks are completed and the entire test is done.... not sure if that's the best approach though?

test/agents/org.js
```diff
@@ -680,4 +680,52 @@ describe('Org', function() {
...
+          orgMock.done();
+          expect(err).to.not.exist();
+        })
+        .then(function(packages) {
+          expect(packages).to.exist();
+          expect(packages.count).to.equal(1);
+          expect(packages.items.length).to.equal(1);
+          expect(packages.items[0].name).to.equal("super-package");
+          done();
+        });
```
rockbot: in this case, because there is no error, `.catch` is definitely not being called, so we're not actually checking if the org mock has completed.

test/agents/org.js
aredridel: It is. However, the bigger issue here is that if those assertions fail, they go nowhere but a rejected promise. We must `.then(function () { done() }, done)` at the end of tests.

jefflembeck:
Tests fixed.

Issue #1602, opened 2015-12-02
Rework subscription deletion to use sub id as key not org name


aredridel:

rockbot:
will need @chrisdickinson's assistance, as well

aredridel:
Awaiting prioritization and coordination in issue-tracker.

PR #1606, opened 2015-12-02
Confirm deleting subscriptions

### LINKED ISSUE #1586: "Delete" X button in orgs list should have a warning dialog
Clicking the X on this page https://www.npmjs.com/settings/billing next to an org immediately cancels the subscription for that org with no confirmation or explanation. There should be one! Wording and UI to @ncawthon, then to team www for implementation.

aredridel:
Fixes #1586

### DIFF
--- a/handlers/org.js
+++ b/handlers/org.js
@@ -300,6 +300,30 @@ exports.updateOrg = function(request, reply) {
   }
 };
 
+exports.deleteOrgConfirm = function(request, reply) {
+  request.customer.getSubscriptions().then(assertSubscriptionExists).then(function() {
+    return reply.view('user/billing-confirm-cancel', {
+      npm_org: request.params.npm_org
+    });
+  }, function(err) {
+    if (err.statusCode == 404) {
+      return reply.view('errors/not-found').code(404);
+    } else {
+      return reply(err)
+    }
+  });
+
+  function assertSubscriptionExists(subscriptions) {
+    if (!subscriptions.filter(function(e) {
+        return e.npm_org == request.params.npm_org
+      }).length) {
+      var err = new Error("Subscription not found");
+      err.statusCode = 404;
+      throw err;
+    }
+  }
+};
+
 exports.deleteOrg = function(request, reply) {
   if (!request.features.org_billing) {
     return reply.redirect('/org');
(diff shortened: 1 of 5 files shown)


handlers/org.js
```diff
@@ -300,6 +300,37 @@ exports.updateOrg = function(request, reply) {
   }
 };
 
+exports.deleteOrgConfirm = function(request, reply) {
+  if (!request.features.org_billing) {
+    return reply.redirect('/org');
+  }
```
rockbot: we can remove the `org_billing` feature check here :-)

aredridel: Heh, I suppose!

handlers/org.js
```diff
@@ -300,6 +300,37 @@ exports.updateOrg = function(request, reply) {
...
+      return reply.view('errors/not-found').code(404);
+    } else {
+      return reply(err)
+    }
+  });
+
+  function selectSubscription(subscriptions) {
+    var sub = subscriptions.filter(function(e) {
+      return e.npm_org == request.params.npm_org
+    })[0];
```
rockbot: a potential edge case: what if there's more than one subscription for the same org?

handlers/org.js
```diff
@@ -310,7 +341,7 @@ exports.deleteOrg = function(request, reply) {
 
   request.customer.getSubscriptions(function(err, subscriptions) {
     if (err) {
-      return replay.view('errors/internal', err);
+      return reply.view('errors/internal', err);
```
rockbot: oooh - good catch!

handlers/org.js
aredridel: In what case would that happen?

And indeed, if that's possible, that's a problem, already present in the existing design, and would be mitigated by fixing #1602

rockbot: sadly, there are already a couple of cases where people have paid for the same org more than once. admittedly, these issues _should_ be fixed in short order (probably by hand, alas), but it's still worth noting that it's a possibility.

aredridel: Yeah. Won't affect this change, since it calls the same endpoint the previous one did, but yeah, we should handle that with #1602

rockbot: sounds good!

Issue #1609, opened 2015-12-03
marky-markdown rendering bugs are not caught


ceejbot:
Here's an example of a package with a readme that marky-markdown cannot handle:

```json
{"time":"2015-12-03T00:41:21.305Z","hostname":"www-10-west.internal.npmjs.com","pid":1605,"level":"info","name":"1449103281278:www-10-west.internal.npmjs.com:1605:ihoyj7sq:36070","message":"52.6.2.64 - 1449103281278:www-10-west.internal.npmjs.com:1605:ihoyj7sq:36070 [3/Dec/2015:00:41:21 +0000] \"GET /package/gekoq \"  - 26ms"}
```
```
Debug: internal, implementation, error
    TypeError: Cannot call method 'match' of null
  at Object.<anonymous> (/mnt/deploys/newww/node_modules/marky-markdown/lib/github.js:24:18)
  at [object Object].exports.each (/mnt/deploys/newww/node_modules/marky-markdown/node_modules/cheerio/lib/api/traversing.js:294:24)
  at module.exports (/mnt/deploys/newww/node_modules/marky-markdown/lib/github.js:13:12)
  at module.exports (/mnt/deploys/newww/node_modules/marky-markdown/index.js:58:7)
  at /mnt/deploys/newww/presenters/package.js:156:18
  at Command.<anonymous> (/mnt/deploys/newww/lib/cache.js:193:5)
  at Command.nrWrappedHandler (/mnt/deploys/newww/node_modules/newrelic/lib/transaction/tracer/index.js:386:19)
  at Command.wrapped (/mnt/deploys/newww/node_modules/newrelic/lib/transaction/tracer/index.js:157:28)
  at Command.wrappedCallback (/mnt/deploys/newww/node_modules/newrelic/lib/transaction/tracer/index.js:438:66)
  at Command.wrapped (/mnt/deploys/newww/node_modules/newrelic/lib/transaction/tracer/index.js:157:28)
  at Command.b [as callback] (domain.js:183:18)
  at RedisClient.return_reply (/mnt/deploys/newww/node_modules/redis/index.js:642:25)
  at /mnt/deploys/newww/node_modules/redis/index.js:307:18
  at process._tickDomainCallback (node.js:492:13)
  at process.wrappedFunction (/mnt/deploys/newww/node_modules/newrelic/lib/transaction/tracer/index.js:271:51)
```

[...]

PR #1610, opened 2015-12-03
bump marky-markdown dependency

bcoe:
By bumping marky-markdown, we fix a 500 error that was being caused by image urls in this format:

`Methods`

I think what the user actually intended was `[Methods](#methods)`, but better that we handle this edge-case without exploding.

## Note

we should test thoroughly because `marky-markdown` `5.4.0` -> `5.5.1` updates quite a few dependencies:

CC: @ceejbot @ashleygwilliams

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "lodash": "^3.8.0",
     "mailchimp-api": "^2.0.7",
     "malarkey": "^1.3.2",
-    "marky-markdown": "~5.4.0",
+    "marky-markdown": "~5.5.1",
     "moment": "^2.10.2",
     "moment-tokens": "^1.0.0",
     "month": "^1.0.0",


bcoe:
see https://github.com/npm/newww/issues/1609

ashleygwilliams:
hey @bcoe ! anything more we need to do to merge/deploy this? happy to help :)

bcoe:
@ashleygwilliams would love the help. We just need to:

`git push origin +bump-marky-markdown:deploy-staging`

Visit a few package pages to ensure that things are still working great, and:

1. merge (after making sure a couple folks on www are fine with this bump).
2. `git push origin +master:deploy-production`
3. we'll see output in `deploys`, once the code is live we should check that READMEs are rendering well in production.

aredridel:
LGTM too.

ashleygwilliams:
cool @bcoe @aredridel, i'm on it!

ashleygwilliams:
staging looks good. merging.

ashleygwilliams:
deployed to prod.

PR #1615, opened 2015-12-04
Restart canceled org

jefflembeck:
Add the ability to restart an organization to the org billing update. This will provide us with the ability to add a "restart" button.


agents/customer.js
```diff
@@ -211,77 +211,91 @@ Customer.prototype.createSubscription = function(planInfo, callback) {
...
+        err = new Error(body);
+        err.statusCode = resp.statusCode;
+        return reject(err);
+      }
 
-    return callback(null, body);
-  });
+      return accept(body);
+    });
+  }).nodeify(callback);
```
rockbot: let's not forget to get rid of `.nodeify(callback)` when we move everything over to promises :-)

jefflembeck: I'm very excited for this refactor. Also, an eventual switch from "wrap everything in a promise"

test/handlers/org.js
```diff
@@ -1890,4 +1890,212 @@ describe('deleting an org', function() {
...
+        })
+        .post("/sponsorship/3839f888")
+        .reply(200, {
+          "id": 268,
+          "npm_user": "betty",
+          "created": "2015-12-03T21:54:17.673Z",
+          "updated": "2015-12-03T21:54:17.673Z",
+          "deleted": null,
+          "verified": true
+        });
```
rockbot: do we really need this many API calls for a single org restart? That seems... scary.

aredridel:
Good call on the headers.

PR #1623, opened 2015-12-08
"/npm" should take you to "/package/npm". Fixes #1618.

### LINKED ISSUE #1618: npmjs.com/npm is 500
I know there's a route in there to move /npm/* to the CMS, but /npm with no trailing slash should still redirect to /package/npm even if we have to special case it :-)

jonathanmarvens:
Not ready for merging. Just pushing it up.


stubbornella:
It looks like very few people actually hit this URL, so I'm going to close this PR. If we see a lot of log traffic in the future, we'll address it.

Issue #1599, opened 2015-12-01
Allow user to pay for Personal or Orgs from subscriptions screen


seldo:
Instead of the extra tab for "sign up for private modules" here:

![screenshot 2015-11-30 16 18 06](https://cloud.githubusercontent.com/assets/185893/11488779/f7c5a5dc-977d-11e5-80df-26b3bc7f59f0.png)

Switch to an empty state like (deliberately terrible wireframe):

![screenshot 2015-12-01 12 16 18](https://cloud.githubusercontent.com/assets/185893/11512928/6fc397a4-9825-11e5-8154-5d01995be6d4.png)

seldo:
Giving to @ncawthon to actually lay this out, decide button names, headers, etc etc etc.

seldo:
Further clarifying:

Clicking the "personal" sign up button takes you to a screen to enter your credit card details if we don't already have them, and then dropping you back into this subscriptions tab. If we already have billing info, just create a subscription and go back to this tab.

Clicking the "org" sign up button takes you to the existing new-org creation flow.

rockbot:
I'll be handling the implementation of this - feel free to assign it to me when you're done with the designs, @ncawthon :-)

(meanwhile I'm going to start with the wireframes)

ncawthon:
I'm blocked on #1587 for the next half-day - I think @seldo's mocks are
good to get going with... sry

nc

On Tue, Dec 1, 2015 at 4:31 PM, Raquel Vélez <notifications@github.com>
wrote:

> I'll be handling the implementation of this - feel free to assign it to me
> when you're done with the designs, @ncawthon <https://github.com/ncawthon>
> :-)
>
> (meanwhile I'm going to start with the wireframes)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/npm/newww/issues/1599#issuecomment-161141149>.
>



-- 
nc

Nick Cawthon
gaugedesign.com
510 542 7364

rockbot:
Possible user states:

Paid for Personal | Paid for Orgs | thing
----|----|----
:white_check_mark: | :white_check_mark:  | Personal info; Org info; Add orgs
:white_check_mark: | :x: | Personal info; Add orgs
:x: | :white_check_mark: | Personal Info; Add Personal (?), Org info; Add orgs
:x: | :x: | Add Personal, Add Orgs

Questions: 
1) Where do we go to add personal packages? If there are no other things being paid for, we can take the user to the credit card form, but then what?
[...]

PR #1629, opened 2015-12-09
1599 subscriptions

rockbot:
this is still a work in progress until all points have been approved.

in particular:
1. is the flow as expected?
2. there are some missing pieces to the UX (see comments in https://github.com/npm/newww/issues/1599); can we ship without them?
3. how's the wording?

cc @ncawthon, @seldo


seldo:
Can we see this on staging? It is very hard to judge UX from a diff.

seldo:
My stalling tactic has failed! I don't have time to review tonight, will tackle in the morning.

ncawthon:
dub dub - would like to fold this in (if it passes), as it addresses some of the missing icon set around the actions tab.

aredridel:
Fold what in?

PR #1631, opened 2015-12-09
Validate username early and bail out

aredridel:


facets/user/show-login.js
```diff
@@ -46,7 +47,10 @@ module.exports = function login(request, reply) {
 
         // User is not above the login attempt threshold, so try to log in...
         User.new(request)
-          .login(request.payload, function(er, user) {
+          .login({
+            name: request.payload.name,
+            password: request.payload.password
+          }, function(er, user) {
```
jonathanmarvens: I like that you make this explicit :smile:.

facets/user/show-login.js
```diff
@@ -23,11 +25,10 @@ module.exports = function login(request, reply) {
   if (request.method === 'post') {
 
     if (!request.payload.name || !request.payload.password) {
-      opts.error = {
-        type: 'missing'
```
jonathanmarvens: Is anything expecting the `type` property to exist?

aredridel: Nope. I checked -- in fact, that this is an object is why the error shows `[object Object]` when you leave out the username or password.

jonathanmarvens: Okay, gotcha.

aredridel: I stand corrected. The !#!@# test asserts on this.

facets/user/show-login.js
ceejbot: whitelist ftw because otherwise who knows what they submitted to the endpoint.

aredridel: Yeessss. this is not Rails, we don't want their buts.

jefflembeck:
:sheep:

PR #1632, opened 2015-12-09
Make sure 404 turns out a not found page

### LINKED ISSUE #1611: Accessing the page for a non-existent organizaton is a redirect loop
We redirect to /org/XXXX?notice= to display the notice about XXSS not existing.

jefflembeck:
A 404 on an org is currently causing a redirect loop.

This fixes that.

Fixes #1611

### DIFF
--- a/handlers/org.js
+++ b/handlers/org.js
@@ -121,7 +121,9 @@ exports.getOrg = function(request, reply) {
     .catch(function(err) {
       request.logger.error(err);
 
-      if (err.statusCode < 500) {
+      if (err.statusCode === 404) {
+        return reply.view('errors/not-found', err).code(404);
+      } else if (err.statusCode < 500) {
         return request.saveNotifications([
           P.reject(err.message)
         ]).then(function(token) {
@@ -133,7 +135,7 @@ exports.getOrg = function(request, reply) {
           request.logger.log(err);
         });
       } else {
-        return reply.view('errors/internal', err);
+        return reply(err);
       }
     });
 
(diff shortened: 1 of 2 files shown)


handlers/org.js
```diff
@@ -121,7 +121,9 @@ exports.getOrg = function(request, reply) {
     .catch(function(err) {
       request.logger.error(err);
 
-      if (err.statusCode < 500) {
+      if (err.statusCode === 404) {
+        return reply.view('errors/not-found', err).code(404);
```
aredridel: Could just return `reply(err)` here

test/handlers/org.js
```diff
@@ -275,24 +275,8 @@ describe('getting an org', function() {
...
-      expect(resp.statusCode).to.equal(302);
-      tokenFacilitator.read(token, {
-        prefix: "notice:"
-      }, function(err, notice) {
-        expect(err).to.not.exist();
-        expect(notice.notices).to.be.array();
-        expect(notice.notices[0]).to.equal('Org not found');
-        done();
-      });
+      expect(resp.statusCode).to.equal(404);
```
aredridel: Shiny

handlers/org.js
rockbot: oooh, yes! (as per https://github.com/npm/newww/blob/master/lib/error-handler.js#L16)

jefflembeck: @aredridel this blows up my tests and gives an error code of 500

jonathanmarvens:
LGTM, @jefflembeck!

PR #1637, opened 2015-12-09
Fix expected error message in addPackages() test

nexdrew:
Change expected error message from `Team or Org or not found` to `Team or Org not found` (one too many `or`s in the first one).

Fixes the `addPackages() returns an error if something goes wrong` test.


aredridel:
Nice catch!

PR #1639, opened 2015-12-09
1587 org billing

jefflembeck:
This implements the payment info page.

Along with that, it implements restarting an org.

There is still some work left with having the user confirm decisions that have to do with money.


agents/customer.js
```diff
@@ -379,3 +393,19 @@ Customer.prototype.removeSponsorship = function(npmUser, licenseId, callback) {
...
+  return self.revokeSponsorship(npmUser, oldLicenseId)
+    .then(function() {
+      return self.extendSponsorship(newLicenseId, npmUser);
+    })
+    .then(function(sponsorship) {
+      return self.acceptSponsorship(sponsorship.verification_key)
+        .catch(function(err) {
+          if (err.statusCode !== 409) {
+            throw err;
+          }
```
rockbot: what happens if there's an error, but it's not a 409?

jefflembeck: A 409 is eaten in our current flow. It is returned when the user is already sponsored. Long term, this could be pointed out in a notice, but it should still be treated as a success

Issue #1501, opened 2015-11-17
Directions for reporting errors are incorrect for npmo mode


aredridel:
It should direct folks to support rather than IRC

rockbot:
cc @ncawthon

nexdrew:
How does one invoke the "directions for reporting errors" in newww?

nexdrew:
## Current

This is how `internal.hbs` currently renders with `FEATURE_NPMO=true`:

![screen shot 2015-12-09 at 10 54 59 am](https://cloud.githubusercontent.com/assets/1929625/11690277/e1be3f8c-9e63-11e5-97ce-60648a994723.png)

## Proposed

I imagine we'll want the error log with stack trace, which only the customer will have access to in an On-Site instance (and which may not be very easy for them to extract). For this reason, I suggest we either:

1. Show the error with stack trace in the UI
2. Provide a link to download the log file (or at least the relevant portion that contains the error with stack)
3. Do both

Perhaps something like this:

![screen shot 2015-12-09 at 12 17 29 pm](https://cloud.githubusercontent.com/assets/1929625/11692701/ff6a94d0-9e6e-11e5-85e6-f51512dd2e30.png)

aredridel:
That makes sense. And directing to more official, enterprise support might be smart. With instructions about how to identify themselves, maybe a license ID? Should probably get @npm/support-team to weigh in what would be useful in diagnostics. We certainly have a reduced need to hide errors from the public site.

Angelfirenze01:
Having needed to send help requests to NPM several times, I agree that the much more explanatory version of the error message helps a programmer with a problem and the programmer helping them sort out what happened much faster.

It’s along the same line as when you send an error report from your hard drive and are asked for supplemental information.

nexdrew:
@aredridel Good point about ID/license info. :+1:

I don't think "more official, enterprise support" exists at the moment. :confused:

Do you think the error page should show the full error/stack trace for the public site too?

aredridel:
No, I think the public site should probably be a bit more conservative -- we are wide wide open to the Internet and we will at least get a visit from chalker if we expose that :)

aredridel:
More official, Enterprise support is support@npmjs.com, or slack if they have that level of support as opposed to IRC

nexdrew:
Closing since main concerns were addressed by #1641. If UI changes are desired, please open another issue.

PR #1641, opened 2015-12-10
Support for npmo-specific error page

nexdrew:
Meant to address #1501.

Things changed:

- Error handling in `handlers/homepage.js` adds a `npmoFullStack` property to the view context
- The `templates/errors/internal.hbs` view renders slightly differently when the `npmoFullStack` context property is defined:
    - Uses `container` instead of `container narrow` class on outer div
    - Displays a slightly different message that displays the support@npmjs.com email and no longer mentions twitter or IRC
    - Displays full stack trace of error instead of just a correlation ID

Things still missing:

- [x] Apply `npmoFullStack` to view context for all npmo-relevant handlers that forward to the `errors/internal` view
- [x] Unit tests for view changes
- [x] Unit tests for handler changes
- [x] ~~Display license info (email address/billing id and license key) - _this data is currently not available in newww_~~ Out of scope
- [x] ~~Potentially provide a link/button to download the newww log file (not a requirement of #1501)~~ Out of scope

Things I don't like about the current implementation (and would like feedback on):

1. In order to provide a "full" stack trace to the user, I copied the private `stackToString` method from `bole` and stuck it in `lib/utils.js`. I figured it was either this or pull in another dependency.
2. Manually checking the `npmo` feature and applying the `npmoFullStack` to the view context in every handler seems clumsy/cumbersome. Is there a way to do this in a central place that automatically gets applied to a rendering of a specific view? I tried `adapters/bonbon.js` and `lib/error-handler.js` but the `stack` from the error seems to be lost by that point (inaccessible from `response.source.context`).

EDIT: I found a way to fix (2) above and centralize the logic in `lib/error-handler.js`, but it means changing handlers and facets to do `reply(err)` instead of `reply.view('errors/internal', err)`. That will be addressed by the first item in the todo list above.

### DIFF
--- a/handlers/access.js
+++ b/handlers/access.js
@@ -27,13 +27,20 @@ module.exports = function(request, reply) {
           reply.view('errors/not-found').code(404);
           break;
         default:
-          reply.view('errors/internal', context).code(500);
+          err.internalStatusCode = 500;
+          reply(err);
       }
       return promise.cancel();
     })
     .then(function(pkg) {
       cpackage = omit(pkg, ['readme', 'versions']);
       return Collaborator(loggedInUser && loggedInUser.name).list(cpackage.name);
+    })
+    .catch(function(err) {
+      request.logger.error('unable to get collaborators for package', request.packageName);
+      request.logger.error(err);
+      reply(err);
+      return promise.cancel();
     })
     .then(function(collaborators) {
       cpackage.collaborators = collaborators;
@@ -71,6 +78,12 @@ module.exports = function(request, reply) {
         return reply.view('package/access', context);
       }
 
+    })
+    .catch(function(err) {
+      request.logger.error('unable to verify access for package', request.packageName);
+      request.logger.error(err);
+      reply(err);
+      return promise.cancel();
     });
 
 };
--- a/handlers/access.js
+++ b/handlers/access.js
@@ -27,7 +27,7 @@ module.exports = function(request, reply) {
           reply.view('errors/not-found').code(404);
           break;
         default:
-          err.internalStatusCode = 500;
+          err.statusCode = 500;
           reply(err);
       }
       return promise.cancel();
--- a/handlers/package.js
+++ b/handlers/package.js
@@ -101,7 +101,8 @@ exports.show = function(request, reply) {
       }
 
       request.logger.error(err);
-      reply.view('errors/internal', context).code(500);
+      err.internalStatusCode = 500;
+      reply(err);
       return;
     });
 };
--- a/handlers/package.js
+++ b/handlers/package.js
@@ -101,7 +101,7 @@ exports.show = function(request, reply) {
       }
 
       request.logger.error(err);
-      err.internalStatusCode = 500;
+      err.statusCode = 500;
       reply(err);
       return;
     });
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -37,4 +37,16 @@ utils.getUserIP = function(request) {
   return request.headers['fastly-client-ip'] || request.headers['x-forwarded-for'] || "unknown";
 };
 
+// stolen from bole private method of same name
+utils.stackToString = function stackToString (err) {
+  var stack = err.stack;
+  var cause;
+
+  if (typeof err.cause === 'function' && (cause = err.cause())) {
+    stack += '\nCaused by: ' + stackToString(cause);
+  }
+
+  return stack;
+};
+
 module.exports = utils;
--- a/services/npme/methods/verifyTrial.js
+++ b/services/npme/methods/verifyTrial.js
@@ -11,6 +11,11 @@ module.exports = function(verificationKey, callback) {
     json: true
   }, function(er, resp, trial) {
 
+    if (er) {
+      log.error('error finding verification key ' + verificationKey, er);
+      return callback(er);
+    }
+
     if (resp.statusCode === 404) {
       log.error('unable to find verification key ' + verificationKey);
 
@@ -31,7 +36,12 @@ module.exports = function(verificationKey, callback) {
     request.put({
       url: trialEndpoint + '/' + trial.id + '/verification',
       json: true
-    }, function(er, resp, verifiedTrial) {
+    }, function(err, resp, verifiedTrial) {
+
+      if (err) {
+        log.error('error starting trial, from hubspot', err);
+        return callback(err);
+      }
 
       if (resp.statusCode === 200) {
         return callback(null, verifiedTrial);
--- a/templates/errors/internal.hbs
+++ b/templates/errors/internal.hbs
@@ -8,7 +8,7 @@
   {{#if npmoFullStack}}
   <p>
     If you feel this error is the result of a bug, please email
-    <a href="mailto:support@npmjs.com">support@npmjs.com</a>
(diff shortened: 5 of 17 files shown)


nexdrew:
Oh, this is what it currently looks like:

screen shot 2015-12-09 at 6 13 47 pm

lib/utils.js
```diff
@@ -37,4 +37,14 @@ utils.getUserIP = function(request) {
...
+// stolen from bole private method
+utils.stackToString = function stackToString (e) {
+  var s = e.stack, ce;
+
+  if (typeof e.cause === 'function' && (ce = e.cause()))
+    s += '\nCaused by: ' + stackToString(ce);
+
+  return s;
+};
+
```
rockbot: We tend to avoid single-letter variable names in this codebase - would you mind changing to something like the following?

```js
var stack = err.stack;
var cause = err.cause;
```

It'll keep things super clear for future code reviews :-)

templates/errors/internal.hbs
```diff
@@ -1,10 +1,19 @@
-<div class="container narrow">
+<div class={{#if npmoFullStack}}"container"{{else}}"container narrow"{{/if}}>
```
rockbot: could simply to `"container{{#unless npmoFullStack}} narrow{{/unless}}"

nexdrew:
@rockbot The logic I need in `bole` is private. :confused: I thought about trying to give `bole` a stream that just writes to a string, but I don't think `bole` lets you do that in a stateless way (e.g. per-request). Plus that seems like major overkill for something so simple.

I'm not real familiar with `bole` though, so feel free to correct me if I'm wrong.

lib/utils.js
nexdrew: Yes, I will change.

templates/errors/internal.hbs
nexdrew: Good point. I wasn't sure if there was a convention to not include hbs expressions within double-quotes (from other examples I saw), but your version is much easier to read. I will change.

handlers/access.js
```diff
@@ -35,6 +36,12 @@ module.exports = function(request, reply) {
       cpackage = omit(pkg, ['readme', 'versions']);
       return Collaborator(loggedInUser && loggedInUser.name).list(cpackage.name);
     })
+    .catch(function(err) {
```
nexdrew: Note that these rejection handlers (new lines 39 and 82) are required to avoid "Unhandled rejection" errors (which kill the HTTP response) if the `Collaborator` agent fails.

services/npme/methods/verifyTrial.js
```diff
@@ -11,6 +11,11 @@ module.exports = function(verificationKey, callback) {
     json: true
   }, function(er, resp, trial) {
 
+    if (er) {
```
nexdrew: These error checks (new lines 14 and 41) are required to avoid potential "Cannot get statusCode of undefined" uncaught errors.

test/mocks/server.js
```diff
@@ -1,7 +1,7 @@
 var path = require('path');
 var Hapi = require('hapi');
 
-module.exports = function(done) {
+module.exports = function(done, additionalPlugins) {
```
nexdrew: The optional `additionalPlugins` argument allows plugins missing from the static `server.register` list (lines 29 - 42) to be applied programmatically for tests that need them.

aredridel:
Stealing from bole++ -- though if there's a dependency that does that, let's just use it. We're not afraid of dependencies if they're not tangly like that.

handlers/access.js
```diff
@@ -35,6 +36,12 @@ module.exports = function(request, reply) {
...
+      request.logger.error('unable to get collaborators for package', request.packageName);
+      request.logger.error(err);
+      reply(err);
+      return promise.cancel();
```
aredridel: TIL!

handlers/package.js
```diff
@@ -101,7 +101,8 @@ exports.show = function(request, reply) {
       }
 
       request.logger.error(err);
-      reply.view('errors/internal', context).code(500);
+      err.internalStatusCode = 500;
```
aredridel: why `internalStatusCode` over `statusCode`?

services/npme/methods/verifyTrial.js
aredridel: \o/

handlers/package.js
nexdrew: Good question. I thought about using `statusCode` but I wanted to preserve the integrity of the original error in logs and in downstream code, in case it already has a `statusCode` that is different than the one we want to return for this error in our response. So to mimic the explicit 500 from prior code, I just added a new property that is given precedence in `lib/error-handler.js`.

(I actually meant to assign `internalStatusCode` *before* logging the error here, just so the log statement represents an accurate picture of what will propagate downstream.)

handlers/access.js
nexdrew: I was really confused by this in my testing. I didn't realize the position of the `catch` affected the scope of the "try" in a Bluebird Promise chain. After trying several things and reading [this](https://github.com/petkaantonov/bluebird/issues/476) and [this](http://stackoverflow.com/questions/26076511/handling-multiple-catches-in-promise-chain), it seemed the easiest thing was to just add more rejection handlers.

Unfortunately the code is not very DRY, but it's at least legible and has a clear intent, I believe.

handlers/package.js
aredridel: How about wrapping the error with a derivative with VError, and setting status code on that?

nexdrew: Sure, that would work. I just thought this was a little lighter weight. I can nuke `internalStatusCode` if you'd prefer.

handlers/access.js
aredridel: Yeah. Making a function to just do `.catch(handleTheThing)` might tighten that up but not bad.

aredridel: The other way to do it is to restructure so that the promises are nested, but that's a more invasive refactor.

nexdrew: So I tried DRYing up the rejection handlers with something like this:

```
function errorHandler(msg, request, reply, promise) {
  return function handleError(err) {
    request.logger.error(msg, request.packageName);
    request.logger.error(err);
    reply(err);
    return promise.cancel();
  };
}
```

And using it like:

```
.catch(errorHandler('unable to get collaborators for package', request, reply, promise))
```

But then I have to split up the chain to maintain appropriate state of the `promise`, which kinda defeats the purpose of a single chain that uses `promise.cancel()` on the first rejection.

So I'd either have to nest or just use a single `catch` at the end, in which case I'd lose context about where in the chain the rejection occurred. This would probably be a good candidate for Bluebird's type- or predicate-based error filtering/catching.

Instead of undertaking a larger refactoring, I'll just leave it as is for now.

handlers/package.js
aredridel: Yeah, I'm torn. I like lightweight, but error tracing is ... well, you don't know you're gonna need it until you do, and then you're glad you had the extra info.

nexdrew: After much ado about nothing, I removed `internalStatusCode` in favor of convention, overriding the `statusCode` instead. See commit 6e6a626.

aredridel: Shiny.

nexdrew:
I think this PR is now complete. :sunglasses:

aredridel:
LGTM!

PR #1647, opened 2015-12-10
fixes bug where unpaid users get 500 instead of billing page

rockbot:
to replicate the bug:

* log in as a user with no payment information
* click on "sign up for private modules" on the profile page
* get 500 

what's expected:
* the user should be taken to the subscriptions page

===

the following PR fixes the bug, but I can't replicate the bug in tests. Any thoughts on how to test this properly?

### DIFF
--- a/handlers/customer.js
+++ b/handlers/customer.js
@@ -36,10 +36,16 @@ customer.getBillingInfo = function(request, reply) {
 
         return {
           customer: customer
+        };
+      }, function(err) {
+        if (err.statusCode === 404) {
+          return {};
+        } else {
+          throw err;
         }
-      })
+      });
     }, function(err) {
-      if (err.statusCode == 404) {
+      if (err.statusCode === 404) {
         return {};
       } else {
         throw err;
--- a/handlers/customer.js
+++ b/handlers/customer.js
@@ -25,38 +25,30 @@ customer.getBillingInfo = function(request, reply) {
     opts.package = request.query.package;
   }
 
-  P.join(
-    request.customer.getById(request.loggedInUser.email).then(function(cust) {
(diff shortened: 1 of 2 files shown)


handlers/customer.js
```diff
@@ -36,10 +36,16 @@ customer.getBillingInfo = function(request, reply) {
...
           customer: customer
+        };
+      }, function(err) {
+        if (err.statusCode === 404) {
+          return {};
+        } else {
+          throw err;
         }
-      })
+      });
```
ceejbot: I would really like to see these functions pulled out & given names for readability reasons. I have been staring at the forest of braces for a couple of minutes now, figuring out what's an error handler etc.

handlers/customer.js
```diff
@@ -78,19 +76,24 @@ customer.getBillingInfo = function(request, reply) {
         privateModules: privateModules,
         orgs: orgs
       };
-    })
-  ).then(function mixInAdditionalOpts(addopts) {
+    });
+
+  P.join(customerStripeData, customerSubscriptions, function(customer, subscriptions) {
```
ceejbot: Each of these two promises (`customerStripeData` and `customerSubscriptions`) represents a call to an external service. These calls might fail or return surprising results (empty, malformatted, not containing required data) etc. Your tests should verify that a) the happy path works, b) the error page is shown when either or both calls fail, c) all errors thrown anywhere in the process are caught instead of crashing the service. The error that originally triggered the look at this method is probably in category (b). Maybe.

jonathanmarvens: Thanks for the suggestions, @ceejbot!

aredridel:
Oh that looks delightful. The 'return null or throw' catcher is a perfect 'exception filter' and exactly what you want. Nice!

PR #1648, opened 2015-12-11
Set up mechanism to prevent page indexing.

ajcrites:
Also block the access pages from search engine indexing

If you Google "parse messy schedule" the first result is the Github repo and the second is the "access" page. The package page itself doesn't show up.

Since `access` is an administrative page without a lot of useful information for non-maintainers it shouldn't be indexed.


aredridel:
Thoughts @npm/dubdub ?

aredridel:
I am quite a fan here, actually. Controlling the indexing to the public set should help make google return more useful results.

PR #1650, opened 2015-12-11
Remove newrelic module

### LINKED ISSUE #1649: Remove New Relic instrumentation
The New Relic trial has expired, and, per @jefflembeck, did not provide the benefits we hoped for.

Since the `newrelic` module requires env vars or config that affect npmo, we either need to remove the module or add the env vars to npmo. After discussing with the team, we decided to just remove the module and its implicit requirements.

nexdrew:
Undo #1417 and fix #1649. Removes the `newrelic` module and any mention of its requirements/resources.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ node_modules/
 .idea
 dump.rdb
 erl_crash.dump
-newrelic_agent.log
 npm-debug.log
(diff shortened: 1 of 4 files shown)


aredridel:
lgtm

PR #1652, opened 2015-12-15
Do not include hiring-container in search view for npmo mode

nexdrew:
Of the four templates that include a `.hiring-container` (for displaying a who's hiring list), only the package search result view (`templates/registry/search.hbs`) neglects the check for `features.npmo`.

This fixes that.


aredridel:
LGTM!

PR #1653, opened 2015-12-15
Give npmo its own default page title based on customer name

nexdrew:
Makes it easier to distinguish between npm On-Site and the public registry site in browser tabs.

Plus it makes npmo that much more customized for our awesome customers!

Public registry tab:

screen shot 2015-12-14 at 11 22 04 pm

New and improved On-Site tab:

screen shot 2015-12-14 at 11 21 27 pm


bcoe:
I like the ♡

aredridel:
<3 it!

PR #1657, opened 2015-12-15
Move to getLicenseForOrg from getLicenseIdForOrg

jefflembeck:
We will frequently be needing to reference parts of the license instead of just the `license_id`. A highlight to this is that we can do this all in one call now, instead of having to make the call and then loop through.

Also:
Cleanup some linting errors on org handler
Update tests

### DIFF
--- a/agents/customer.js
+++ b/agents/customer.js
@@ -231,32 +231,42 @@ Customer.prototype.cancelSubscription = function(subscriptionId, callback) {
   });
 };
 
-Customer.prototype.getLicenseIdForOrg = function(orgName, callback) {
+Customer.prototype.getLicenseForOrg = function(orgName, callback) {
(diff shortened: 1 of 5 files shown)


agents/customer.js
```diff
@@ -231,32 +231,42 @@ Customer.prototype.cancelSubscription = function(subscriptionId, callback) {
...
       }
 
-      return accept(org.license_id);
+      if (!body.length) {
+        err = new Error('No license for org ' + orgName + ' found');
+        err.statusCode = 404;
+        return reject(err);
+      }
+
+      return accept(body[0]);
```
rockbot: what happens if `body.length` > 1? Is that even possible?

jefflembeck: Not in this case, but without the query string, you can get multiple licenses, so the API remains the same (returning an array)

PR #1663, opened 2015-12-15
Refactor revoke/extend/accept with error handler

jefflembeck:
Catch the 409, also some name changes.


jefflembeck:
Yes. Also, we will eventually rip this out and have one call that won't potentially cause _issues_

aredridel:
LGTM

jefflembeck:
This is not currently a thing. We can make it a thing later, but the order that is currently in place for this process is important. Whatever we can do to make this work should likely be done when we can figure out transactionality at the registry level.

PR #1665, opened 2015-12-16
Add custom tutorial for each package

aredridel:


jonathanmarvens:
@aredridel: Can you push to staging, please?

jonathanmarvens:
Oops, looks like @rockbot is a mind reader.

aredridel:
1, I think makes sense, but I actually think we can push this to prod and refine, because I want @ashleygwilliams input.

2, I agree. I can just do that.

3. Yes indeed, let's push and refine.

aredridel:
Discussed in triage: going to ship, then refine with @ashleygwilliams input. Will sync up with her.

PR #1666, opened 2015-12-16
adds classes to org nav to appease the newww-tests deity

rockbot:
there's a bug in the u2o newww-tests test, in which we need to click on the `members` tab to see the members in the org... but in order to click on the tab, we need some sort of defining characteristic; a class will do just the trick.


aredridel:
Works for now! Makes me want to invent a whole taxonomy but this will do.

PR #1667, opened 2015-12-16
Validate phone numbers on the enterprise sign up page. Fixes #1447.

### LINKED ISSUE #1447: Improve phone number validation on enterprise signup page
All we do right now is check that it's not empty:

https://github.com/npm/newww/blob/master/facets/enterprise/show-ula.js#L14

I'm sure a client-side regex can be found to validate something vaguely phone-numbery.

jonathanmarvens:


jonathanmarvens:
__CC:__ @npm/dubdub, @seldo.

facets/enterprise/show-ula.js
```diff
@@ -23,9 +24,18 @@ module.exports = function createHubspotLead(request, reply) {
   };
 
   var validatedData = Joi.validate(request.payload, schema, joiOptions);
+  var phoneNumberDataValid = googleLibphonenumber.PhoneNumberUtil.isViablePhoneNumber(request.payload.phone);
 
-  if (validatedData.error) {
+  if (validatedData.error || (! phoneNumberDataValid)) {
```
aredridel: don't need the extra parens, but it's pretty clear either way

facets/enterprise/show-ula.js
```diff
@@ -23,9 +24,18 @@ module.exports = function createHubspotLead(request, reply) {
...
     opts.errors = validatedData.error.details;
```
aredridel: This could be undefined, couldn't it, if only the phone number fails?

facets/enterprise/show-ula.js
jonathanmarvens: Yes, I know. That's more of a style thing for me. I think it's more readable like this.

aredridel: Works for me then!

facets/enterprise/show-ula.js
jonathanmarvens: I don't *think* so, because this library throws exceptions when it fails. (Maybe I should handle the possible exceptions though!)

aredridel: Definitely! But it is checking for `validatedData.error` or the phone number; if there's no errors, except the phone number, it won't be an array and you can't push the phone number error to it, right?

jonathanmarvens: You're right, @aredridel! Thanks for spotting that.

jonathanmarvens:
@npm/dubdub: Okay to merge?

aredridel:
Almost. That's still got one deref of undefined. If there's no joi error, but there is a phone error, `error.details` will throw.

jonathanmarvens:
@npm/dubdub: Okay to merge now :smile:?

aredridel:
Works for me! Nice!

aredridel:
:sheep: it!

PR #1671, opened 2015-12-17
ensures charset=utf-8 on response headers (fixes #1301)

### LINKED ISSUE #1301: Send charset in `content-type` http headers.
Now: `content-type:text/html`. Should be something like: `Content-Type: text/html; charset=utf-8`.
See http://www.w3.org/International/O-HTTP-charset.en.php.

This is not critical for modern browsers, but it was critical for older browsers (like ie8, I don't remember exact versions and don't have them to check). Also note that you have user-controlled content just above the `<meta charset="utf-8">`, in `<title>`.

_I was told to report this issue here instead of a private chat in irc._

rockbot:
also reordered the lifecycle extensions to be [in order](http://hapijs.com/api#request-lifecycle).

### DIFF
--- a/adapters/bonbon.js
+++ b/adapters/bonbon.js
@@ -104,6 +104,10 @@ exports.register = function(server, options, next) {
       case "view":
         request.response.source.context = Hoek.applyToDefaults(options, request.response.source.context);
         request.response.source.context.user = request.loggedInUser;
+
+        // appends charset to content-type header, for security reasons
+        request.response.type('text/html').charset('utf-8');
+
         break;
       case "plain":
         if (typeof (request.response.source) === "object") {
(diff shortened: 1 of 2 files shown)


adapters/bonbon.js
```diff
@@ -110,22 +127,8 @@ exports.register = function(server, options, next) {
...
-      value: latency,
-      page: request.timing.page,
-    });
-
-    // TODO log request info in as close to common log format as possible
-    request.logger.info(toCommonLogFormat(request, {
-      ipHeader: 'fastly-client-ip'
-    }), latency + 'ms');
+    // appends charset to content-type header, for security reasons
+    request.response.type('text/html').charset('utf-8');
```
aredridel: I think there's a conditional you can use here -- to see if it should be html or what.

aredridel: And there's also some messing with content types going on in the `vision` module

jefflembeck:
:sheep:

PR #1675, opened 2015-12-17
adds updated stripe checkout api to connect-src in csp

rockbot:


jefflembeck:
:sheep: when travis is done

PR #1677, opened 2015-12-18
Turn on esnext for jshint-based tooling

jefflembeck:
Hey Hey, we have node 4 now! Updating jshintrc so my text editor doesn't yell at me for using es2015 stuff.


aredridel:
What features does `esnext` turn on? node 4 only supports some.

jefflembeck:
Just found out that `esnext` is deprecated, need to switch to `esversion` or fix my local editor to stop blowing up :D

PR #1683, opened 2015-12-21
Payment info

jefflembeck:
Lo and behold, the payment info tab.


handlers/customer.js
```diff
@@ -54,36 +54,31 @@ customer.getBillingInfo = function(request, reply) {
       request.logger.error('unable to get subscriptions for ' + request.loggedInUser.name);
       request.logger.error(err);
       return [];
-    })
-    .then(function(subscriptions) {
+    }).then(function(subs) {
 
-      var subs = subscriptions.filter(function(sub) {
-        return sub.status === "active";
-      });
```
rockbot: do we no longer need this status piece?

jefflembeck: Yeah, I misread originally. We're supposed to show all of them.

handlers/org.js
```diff
@@ -4,8 +4,13 @@ var User = require('../models/user');
 var P = require('bluebird');
 var Joi = require('joi');
 var invalidUserName = require('npm-user-validate').username;
+var _ = require('lodash');
+var moment = require('moment');
 var URL = require('url');
 
+const defaultPricePerOrgUser = 700;
+const startingPriceForOrg = 14;
```
rockbot: are these values that we want to have hard-coded? 

if so, i'd prefer the semantic ALL_CAPS for constants (from C++ land, lolfunwoo)

jefflembeck: I can get down with that.

Do we want them hard-coded? Woof. ENV var?

handlers/customer.js
rockbot: ah, ok!

handlers/org.js
```diff
@@ -4,8 +4,12 @@ var User = require('../models/user');
 var P = require('bluebird');
 var Joi = require('joi');
 var invalidUserName = require('npm-user-validate').username;
+var _ = require('lodash');
+var moment = require('moment');
 var URL = require('url');
 
+const STARTING_PRICE_FOR_ORG = process.env.ORG_STARTING_PRICE || 14;
+
```
rockbot: :+1: thank you :-)

handlers/org.js
aredridel: Not an env var. This is not something with a lifecycle that matches them. It should be hard-coded in the code like this. Other code changes are required to support that, it's not a knob that can just be turned, so it's not a thing that should be a knob.

jefflembeck:
@rockbot `isPaidSuperAdmin` should not affect the pages you're able to visit, but _will_ affect whether or not you're given the UI for making changes to an org.

aredridel:
This sure looks good.

PR #1684, opened 2015-12-21
1681 nav flexbox

ncawthon:
- didn't flexbox it; just removed container and added padding - figured we'll rewrite soon
- saw some extraneous code around the feature box, removed that
- do we want to keep `nav` elements in the `header.sytl`, or move it to `nav.styl`?


aredridel:
The tests for this failed -- can you rebase it and try against current master?

PR #1692, opened 2015-12-22
Removed the honeypot from the sign in page.

### LINKED ISSUE #1674: Users get an unclear error when the honeypot is filled for */login*. (As a "400 Bad Request" response.)
@npm/dubdub:
We've been receiving a lot of support requests that basically point to this issue.
We should fix this.

Related to #694.

jonathanmarvens:
Fixes #1674.


jonathanmarvens:
__CC:__ @aredridel.

aredridel:
Elegant and minimal! No repetition! :sheep: it!

PR #1700, opened 2015-12-29
A couple small CSP tweaks

aredridel:


handlers/ops.js
```diff
@@ -22,8 +22,10 @@ module.exports = {
   },
 
   csplog: function(request, reply) {
-    request.logger.warn('csp report');
-    request.logger.warn(request.payload);
+    if (!(request.payload && request.payload['csp-report'] && /^safari-extension:/.test(request.payload['csp-report']['source-file']))) {
```
ceejbot: That should shut up some noise.

aredridel: That was the goal. So tired of seeing CSP spam on every !@#!@# page. Makes it hard to notice real errors.

PR #1714, opened 2016-01-05
Stop caching templates in dev, avoiding need for server restart while editing

aredridel:
cc: @jefflembeck


jefflembeck:
This is :fire:

:sheep:

PR #1715, opened 2016-01-05
Marketing sidebar blob

isaacs:
Sort of like the CMS-driven upsell box, but way smaller, and fewer features.

Deployamating to stage coach presently.

### DIFF
--- a/assets/styles/layout.styl
+++ b/assets/styles/layout.styl
@@ -384,6 +384,8 @@ ul.humans
 
 .ad, .marketing
   margin-left -10px
+  @media only screen and (max-width: 929px)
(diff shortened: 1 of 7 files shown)


bcoe:
can we feature flag this so that it doesn't show up in npm On-Site? or better still, is there something useful it could show in On-Site?

isaacs:
Feature-flagging out of on-site is a great idea. :+1:

aredridel:
LGTM pending feature flagging.

aredridel:
LGTM!

bcoe:
thanks for the npmo check \o/

Issue #1708, opened 2016-01-04
Restart paying for an org when you are a current customer who did not cancel after org was created the first time


jefflembeck:
The backbone for this is already in place.

User who cancelled their org, realizes they made a mistake:

1. Go to the *payment info page* for the org
1. Click restart button
1. Success banner is shown saying "Congrats, you've reinstated your org"

or 

1. Go to the the *billing page* for the user
1. Click restart button
1. Success banner is shown saying "Congrats, you've reinstated your org"

NB: there are two pages they can reinstate the org from

Implementation details:
All we should need is a button on the org payment info page that switches when the org is canceled, which is already on the object we're sending the the template - https://github.com/npm/newww/blob/master/handlers/org.js#L159

Restarting the org is a matter of posting to `/org/{org}` with an `updateType` in the payload of `restartOrg`.

stubbornella:
@ncawthon - please change the wording and add design comps

jefflembeck:
Issue resolved and in production

PR #1718, opened 2016-01-06
Add restart button to payment info page

jefflembeck:
This restarts payment on an organization for a customer that has not
deleted their account or the license for this (deleted by stopping
payment entirely)

Addresses #1708


jefflembeck:
No unit test, just adding markup.

We likely can get a good selenium set of tests once we have the rest of restart working.

PR #1721, opened 2016-01-06
WIP: Add context to the unexpected status code exception in lib/cache

aredridel:
Needs tests to be adjusted to check for err.statusCode instead of matching the message. /cc @jonathanmarvens


stubbornella:
Is this ready to go?

ashleygwilliams:
@stubbornella looks like it is waiting for a review @jonathanmarvens .

 i came just to say yay WIP PRs, a fave workflow of mine :) (tho fwiw labels help because they are easier to remember to change than the subject line)

PR #1724, opened 2016-01-07
fix(CSS): change imports order since colors are needed by variables

Delapouite:
Hi

variables.styl needs the color `npmred`
https://github.com/npm/newww/blob/master/assets/styles/variables.styl#L54

reset.styl also need colors to define the `hr` element 
https://github.com/npm/newww/blob/master/assets/styles/reset.styl#L27

and the `body` element
https://github.com/npm/newww/blob/master/assets/styles/reset.styl#L74-L75

Therefore, colors.styl should be imported first.

### DIFF
--- a/assets/styles/index.styl
+++ b/assets/styles/index.styl
@@ -1,6 +1,8 @@
+// variables and reset depend on colors, so it should be imported first
+
+@import "colors"
 @import "variables"
 @import "reset"
-@import "colors"
 @import "typography"
 @import "icons"
 @import "iconic"


aredridel:
Is this not resolved at compile time, regardless of order? Or is the actually introducing changes to the output

Delapouite:
Thanks for having a look at this PR.

If you search in the current CSS file deployed on the website for `npmred`: 
https://www.npmjs.com/static/css/index.7e101822.css

```css
.red-on-hover:hover,
.sidebar a:hover,
.package-widget a.name:hover,
.package-widget .quiet a:hover,
.ad a.quiet:hover {
  color: npmred;
}
```

The variable has not been replaced. The order of `import` is important.

aredridel:
Aha! That is important then!

aredridel:
LGTM then! Always nice to see what problem is being solved in the PR.

aredridel:
Thank you!

PR #1725, opened 2016-01-07
Fix coupon codes and straighten out updating billing info vs private module subscriptions even more

aredridel:
The old conflation of updating billing info with private module subs was just unfortunate. This untangles it and directs everything through specifically subscription or update flows, rather than ever mixing them.


aredridel:
Okay, I have a feeling here: why hide this form? It's got a single purpose now, this page has had its other features almost unloaded except the link to cancel. How about removing the link to hide-show and always showing it if you're a customer.

jefflembeck:
:+1: :sheep:

PR #1728, opened 2016-01-08
Add newww tests to newww

rockbot:
The problem: 
* in order to write integration tests, we have (to remember) to write them in a completely different repo.
* in order to run those same integration tests, we have to go to a completely different repo and _run them by hand_
* our integration tests are rather paltry compared to our unit tests

The (proposed) solution:
* move them into newww, so everything's in the same repo (and so we can easily write the tests as we write code)
* have them run _automatically_ when we push to staging (and only when we push to staging), with recordings and screenshots saved in browserstack

Already-identified problems:
* there may still be flakiness with browserstack, which we'll find the first few times... and we can always turn it off in .travis.yml as needed.
* as per ^, we've got unbelievable timeouts for our tests (240 seconds). I'd love to bring that number down, if we can figure out what's going on with the flakiness.
* our browserstack account may need to be upgraded (we should be able to assess whether it's worth paying for before our current number of minutes gets used up)

Any/all comments welcome and encouraged.


seldo:
\o/

stubbornella:
@rockbot - where is this PR at? Can you fix the merge conflicts? Danke.

.travis.yml
```diff
@@ -18,3 +18,7 @@ addons:
     packages:
     - gcc-4.8
     - g++-4.8
+script:
+  - if [ "$TRAVIS_BRANCH" == "deploy-staging" ];
+      then npm run all-tests;
+    fi
```
jonathanmarvens: @rockbot I'd put all of this in one line and change `==` to `=`.

__Before:__
```yaml
  - if [ "$TRAVIS_BRANCH" == "deploy-staging" ];
      then npm run all-tests;
    fi
```

__After:__
```yaml
  - if [ "$TRAVIS_BRANCH" = "deploy-staging" ]; then npm run all-tests; fi
```

stubbornella:
Sweet, looks good. @jefflembeck can you take a look?

jefflembeck:
:sheep:

Issue #1730, opened 2016-01-08
A `<div>` in the description of the package breaks the appearance of the project page


lovasoa:
I have the string `<div>` in the description of one of my packages ([react-contenteditable](https://www.npmjs.com/package/react-contenteditable)), and it breaks the appearance of the project page.

See the [`package.json`](https://github.com/lovasoa/react-contenteditable/blob/master/package.json) of the project.

aredridel:
I'm not seeing the break -- what part is broken for you?

lovasoa:
![This is how it is rendered](https://cloud.githubusercontent.com/assets/552629/12207963/d2bad464-b65a-11e5-9142-07d417de0bf6.png)

aredridel:
Oh, I see! Makes sense. Alright! We've had a few README issues, I'm going to file it to solve with those.

lovasoa:
If you look at the html, you will see `<p class="package-description">React component representing a </p>` whereas you should see `<p class="package-description">React component representing a &lt;div&gt; with editable contents</p>`

lovasoa:
And it should look like that
![capture du 2016-01-08 22-58-59](https://cloud.githubusercontent.com/assets/552629/12208059/72d754cc-b65b-11e5-8160-91536817761a.png)

lovasoa:
The problem is that the html is not escaped correctly.

aredridel:
Yup!

aredridel:
This is fixed in production now!

PR #1731, opened 2016-01-08
Use handlebars {{ instead of {{{ to escape html on package description

briandela:
@aredridel I believe that this should fix https://github.com/npm/newww/issues/1730

I couldn't figure out how to run the site locally (I think it's still not possible) so I wasn't able to check for any unintended side effects. I don't think there would be as the only relevant difference between `{{{` and `{{` is that the latter escapes html and the former doesn't.

### DIFF
--- a/templates/package/access.hbs
+++ b/templates/package/access.hbs
@@ -2,7 +2,7 @@
   <h1 class="package-name centered">
     <a href="/package/{{name}}">{{name}}</a>
   </h1>
-  <p class="package-description centered">{{{description}}}</p>
+  <p class="package-description centered">{{description}}</p>
 {{/with}}
 
 {{> package-header package}}
--- a/templates/package/show.hbs
+++ b/templates/package/show.hbs
@@ -10,7 +10,7 @@
       <i class="icon-public"></i>
     {{/if}}
   </h1>
-  <p class="package-description">{{{description}}}</p>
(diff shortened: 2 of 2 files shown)


briandela:
This should fix the issue you see @lovasoa.

lovasoa:
But html is already partially escaped. This is not a security problem, but a feature. Some packages may use that to write something in bold in there package description, for instance. Is that acceptable to remove that feature ?

briandela:
It would be a decision for the npm staff. They could decide to not allow html in there and allow a subset of markdown for bold/italic etc.

That would be a simple handlebars helper to expose that functionality. 

@aredridel if you want, let me know what desired behaviour you would like and I'll update the PR.

briandela:
From my personal perspective, I think it's a little strange that display html would be included in the description `package.json` for a module. The `package.json` description field is used by many tools (e.g. the `npm search` command, rendered in browsenpm.org different than npmjs.org, etc.) and making assumptions about how and were it will be rendered as html feels wrong to be honest.

briandela:
This change also makes the show/access pages consistent with the `package-widget` partial used in multiple other places across the site: https://github.com/npm/newww/blob/master/templates/partials/package-widget.hbs#L14

stubbornella:
Thanks!

stubbornella:
@aredridel - can you test this on staging please?

PR #1735, opened 2016-01-11
restart paying canceled current customer

### LINKED ISSUE #1719: Restart paying with a customer who has restarted
If somebody stops paying us, but then starts paying us again, and then wants to restart an org that was created before they stopped paying us, this should be able to happen.

1. Paid user clicks "cancel"
2. Realizes next month they made a mistake, and wants to restart
3. User visits settings/billing (linked from profile)
4. User chooses to pay for Private Modules
5. Enter credit card info
6. User visits settings/billing (linked from profile)
7. Sees a list of cancelled orgs with a button beside them saying "restart"
8. Click restart
9. Redirected to subscriptions tag (settings/billing) now with the org moved into the active section
10. Show success banner "you successfully reinstated @orgname."

Here's how this needs to work:

Go to restart org
-> License not found
-> Get customer info
-> Get org users from user-acl
-> Create new org license for said customer
-> Attempt to sponsor each user.

This issue does not include the following work:

    Visits cancelled org page - sees call to reinstate it
    We ultimately need to stop really deleting users, but @chrisdickinson is too busy for now

jefflembeck:
Fixes #1719

This adds the functionality of restarting an organization by a customer who had previously stopped paying us while the org was active, but then restarted their account and now wants to restart this org (so the org exists, but is without a license).


aredridel:
:sheep:

PR #1739, opened 2016-01-12
Don't call auth tokens one-time tokens

isaacs:
Auth tokens generated via a cli log-in are not technically one-time
tokens.  They are multi-use tokens.

Very minor copy edit.  Not urgent.


aredridel:
LGTM

PR #1740, opened 2016-01-12
Revert part of commit that hides cancel payment

jefflembeck:
Revert "Unambiguously stop hiding the update payment info screen if it applies"

This partially reverts commit 610e8ee374c7c1f328fb43467d23881cf8dbdd15.


aredridel:
LGTM!

PR #1747, opened 2016-01-14
Use verror and make logging consistent in reset password and password change flows

aredridel:
This may help diagnose some of the user and password reset errors we've seen

cc: @jonathanmarvens


jonathanmarvens:
LGTM, @aredridel. TY!

jonathanmarvens:
Let's :sheep: it!

PR #1748, opened 2016-01-14
Improve license tests to use new nock

aredridel:


jonathanmarvens:
Awesome. LGTM :smile:!

PR #1751, opened 2016-01-14
Remove the form honeypot stuffs entirely.

jonathanmarvens:


aredridel:
AWesome. :sheep: !

Could also change form_csrf to form_security again and just simplify that, but also would be great as a second PR

jonathanmarvens:
@aredridel: Yes, I'm already on that. Separate PR.

PR #1754, opened 2016-01-15
Do not ever just copy things blindly into the json responses

aredridel:
This avoids copying internal data into json responses like it does for the template context. The template context is private but json responses are not.


stubbornella:
@jefflembeck @jonathanmarvens - code review please!

adapters/bonbon.js
```diff
@@ -94,20 +98,12 @@ exports.register = function(server, options, next) {
...
-        // appends charset to content-type header, for security reasons
-        request.response.type('text/html').charset('utf-8');
+    if (request.response.variety == 'view') {
+      request.response.source.context = Hoek.applyToDefaults(options, request.response.source.context);
+      request.response.source.context.user = request.loggedInUser;
 
-        break;
-      case "plain":
-        if (typeof (request.response.source) === "object") {
-          request.response.source = Hoek.applyToDefaults(options, request.response.source);
```
jonathanmarvens: Is this really not needed, @aredridel?

aredridel: Yeah, it's not actually used. Shocked me to find, but just not needed.

jonathanmarvens:
LGTM, @aredridel.

PR #1758, opened 2016-01-16
added missed background color for form inputs

unclechu:
You can see how it is with dark system theme without background color:
selection_001
selection_002


aredridel:
LGTM!

PR #1761, opened 2016-01-19
Add assertions for expanded email data to facet or handler tests that send email

### LINKED ISSUE #1710: Add tests for all outbound emails
Specifically, we need tests that assert on desired output of all used email templates. The tests need to be defined in newww because it's the only place that combines the use of the following 3 decoupled parts:

1. [email templates/views](https://github.com/npm/email-templates)
2. [mailer and template expander](https://github.com/npm/mustache-mailer)
3. consumer-provided data (this repo)

Note that each project has its own tests, which guarantee output of implicit contracts, but there are currently no tests to guarantee that the contracts are not being violated in practice, making it difficult to check for regression when any of the moving parts are changed. For an example, see npm/mustache-mailer#5 and the related discussion of its implications.

It looks like there is an [existing minimal test for the `confirm-user-email` template](https://github.com/npm/newww/blob/v4.0.1/test/services/email.js#L72-L83), but these tests will probably need to be one more layer up (testing consumers of `server.methods.email.send`) in order to include actual consumer-provided data (3rd moving part above) from the facets/handlers.

nexdrew:
Fixes #1710.

This is a precursor to npm/mustache-mailer#5 to guarantee that upstream dependency changes won't result in regression of email data and content sent by the app.

Note that these assertions are correct for existing functionality, even if some results are not as intended. These will be fixed after `mustache-mailer` is updated.

### DIFF
--- a/test/company/contact-support.js
+++ b/test/company/contact-support.js
@@ -3,24 +3,55 @@ var Code = require('code'),
   lab = exports.lab = Lab.script(),
   describe = lab.experiment,
   before = lab.before,
+  afterEach = lab.afterEach,
   after = lab.after,
   it = lab.test,
   expect = Code.expect,
   generateCrumb = require("../handlers/crumb.js"),
+  emailMock,
   server;
 
 before(function(done) {
   require('../mocks/server')(function(obj) {
     server = obj;
+    emailMock = server.methods.email.send.mailConfig.mailTransportModule;
     server.app.cache._cache.connection.client = {};
     done();
   });
+});
+
+afterEach(function(done) {
+  emailMock.sentMail = [];
+  done();
 });
 
 after(function(done) {
   delete server.app.cache._cache.connection.client;
   server.stop(done);
 });
+
+function assertEmail (opts, expectedTo) {
+  var expectedFrom = 'website@npmjs.com';
+  var expectedText = opts.payload.message;
+  var expectedSubject = opts.payload.subject + ' - FROM: "' +
+    opts.payload.name + '" <' + opts.payload.email + '>';
+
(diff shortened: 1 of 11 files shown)


test/company/contact-support.js
```diff
@@ -3,25 +3,58 @@ var Code = require('code'),
...
   generateCrumb = require("../handlers/crumb.js"),
+  MockTransport = require('nodemailer-mock-transport'),
+  emailMock = new MockTransport(),
   server;
 
 before(function(done) {
   require('../mocks/server')(function(obj) {
     server = obj;
     server.app.cache._cache.connection.client = {};
+    server.methods.email.send = require('../../services/email/methods/send');
```
aredridel: Oh this fills me with such rage and loathing.

Maybe it should be in mocks/server. 

Or maybe I should kill it with fire.

aredridel:
Looks good to me. Pushing the mail setup into mocks/server is probably a good plan.

nexdrew:
I'll see what I can do, but there is at least 1 test that depends on the fake rejection/error from `methods.email.send` in `test/mocks/server-methods.js`, so that still needs to be included somehow.

nexdrew:
So I moved the method override and mock email transport into `test/mocks/server-methods.js`, which cleans things up a little, but reading from and resetting the mock's state is still coupled in each test that needs it. Not sure if this is really any better, but it at least keeps the mock defined in a central place and results in one less `require()` statement in tests.

aredridel:
Good enough! That thing is super weird and needs to change but not yet!

nexdrew:
@aredridel Thanks for reviewing this. Ok to merge?

PR #1762, opened 2016-01-19
This fixes a possible viewing of the context' json

jefflembeck:
Users with names that are properties (lower-case only) on Object would
pass this test and be able to see the json we pass to the context. Using
`hasOwnProperty` instead of `in` fixes this


aredridel:
:sheep:!

PR #1763, opened 2016-01-19
Fix drifting code bg when in overflowing pre

remy:
I'm not sure if this has been raised separately, but it's been bugging me for some time and I wanted to send a PR to fix it.

When there's an overflowing `pre` tag with `code` inside, the grey background drifts with the scroll, as per:

before mp4

This change fixes the `code` element with a change to `display: inline-block` and `min-width: 100%` to ensure it stretches across both the width of the entire page (to emulate the `block` style) and the width of the overflowing `pre`, as per:

after mp4

Note that I took care to keep the scrollbar in it's position (with the white background), otherwise I think the CSS could be simplified (maybe).

Anyway... keep on doing the great stuff :heart:

### DIFF
--- a/assets/styles/typography.styl
+++ b/assets/styles/typography.styl
@@ -173,7 +173,8 @@ code
 pre
   overflow-x auto
   code
-    display block
+    display inline-block
+    min-width 100%
     margin 0 0 15px 0
     padding 8px
     border-radius 2px


remy:
Oh, I meant to add, I wasn't sure if there was any tests that I could update for this...so it doesn't include any tests...sorry!

aredridel:
OMG! This is the best kind of contribution!

LGTM!

jonathanmarvens:
Thank you, @remy! LGTM!

Issue #1769, opened 2016-01-20
error when attempting to change account email


snopeks:
Hi team, I tried to change my email on my snopeks account twice just now, got an error each time. Change did not go through. 

Here's the first error number: 1453314531091:www-11-west.internal.npmjs.com:12066:ijlvsav9:51687
Here's the second: 1453314837600:www-12-west.internal.npmjs.com:3902:ijlvs7fg:72313

Lemme know if you need more details. Thanks!

PR #1775, opened 2016-01-20
Include more information in error report saving profile

aredridel:
This should get more diagnostics for #1769.


jonathanmarvens:
LGTM, @aredridel!

Issue #1709, opened 2016-01-04
Restart paying with a non-customer


jefflembeck:
If somebody stops paying us, but then wants to start paying us again by restarting an org. That should definitely be an option!

1. Paid user clicks "cancel" 
1. Realizes next month they made a mistake, and wants to restart
1. User visits settings/billing (linked from profile)
1. Sees a list of cancelled orgs with a button beside them saying "restart"
1. Click restart
1. Enter credit card info
1. Redirected to subscriptions tag (settings/billing) now with the org moved into the active section
1. Show success banner "you successfully reinstated @orgname."

Here's how this needs to work:

Go to restart org
-> Customer is not found
-> Get Billing Details
-> Create new customer
-> Create new org license for said customer
-> Attempt to sponsor each user.

This issue does not include the following work: 

* Visits cancelled org page - sees call to reinstate it 
* We ultimately need to stop really deleting users, but @chrisdickinson is too busy for now (EDIT: stripe makes this impossible - if we get rid of billing info, we get rid of the licensed user)

jefflembeck:
Issues that are related: https://github.com/npm/license-api/issues/101
https://github.com/npm/newww/issues/1587#issuecomment-162998485

stubbornella:
@chrisdickinson - does this issue description match your expectations?

stubbornella:
@ncawthon - can you please insert designs that match from the billing information tab?

stubbornella:
@ncawthon - what wording would you use to link the person to billing information?

chrisdickinson:
> We ultimately need to stop really deleting users, but @chrisdickinson is too busy for now

Responded here: https://github.com/npm/license-api/issues/100#issuecomment-168856163 — We should not plan to have license-api not really delete customers on delete requests.

> Attempt to sponsor each user.

This could be drawn from the sponsorship list for the old license, or alternatively from the list of members of the org. Do we have a preference on which it's drawn from?

jefflembeck:
@chrisdickinson for the first one, yeah, totally impossible, we can work around that
for the second one, I think the latter will do just fine

jefflembeck:
Issue resolved and in production

PR #1776, opened 2016-01-20
1709 restart paying non customer

jefflembeck:
This addresses #1709 

If somebody stops paying us, all together, this allows for that non-customer to subscribe again via restarting an org.


handlers/customer.js
```diff
@@ -319,43 +327,110 @@ customer.subscribe = function(request, reply) {
...
-              if (typeof subscription === 'string') {
-                request.logger.info("created subscription: ", planInfo);
+          return Customer(loggedInUser).createSubscription(planInfo);
+        })
+        .tap(function(subscription) {
+          if (typeof subscription === 'string') {
+            request.logger.info("created subscription: ", planInfo);
+          }
+          var dropCache = P.promisify(User.new(request).dropCache);
+          return dropCache(loggedInUser);
```
jonathanmarvens: @jefflembeck: This isn't required, but to be safe, if I was writing this, I'd write it like this:

```javascript
var user = User.new(request);
var dropCache = P.promisify(user.dropCache, user);

return dropCache(loggedInUser);
```

agents/customer.js
```diff
@@ -67,8 +70,10 @@ Customer.prototype.getStripeData = function(callback) {
       }
 
       if (resp.statusCode === 404) {
-        err = Error('customer not found: ' + self.name);
-        err.statusCode = resp.statusCode;
+        err = Object.assign(new Error('Customer not found: ' + self.name), {
+          code: 'ENOCUSTOMER',
+          statusCode: 404
```
jonathanmarvens: @jefflembeck: Why not `statusCode: resp.statusCode`? I feel like that's a little bit cleaner. Obviously no biggie though.

agents/customer.js
```diff
@@ -249,8 +254,11 @@ Customer.prototype.getLicenseForOrg = function(orgName, callback) {
       }
 
       if (resp.statusCode === 404) {
-        err = new Error('Customer not found');
-        err.statusCode = 404;
+        err = Object.assign(new Error('Customer not found'), {
+          code: 'ENOCUSTOMER',
+          statusCode: 404
```
jonathanmarvens: Same as https://github.com/npm/newww/pull/1776/files#r50441374 ...

handlers/customer.js
jefflembeck: Word. We have an issue in for promisifying `dropCache`, so I think it'll wait until then.

agents/customer.js
jefflembeck: definitely cleaner, but I think I just wanted it to read super explicitly.

aredridel:
I think this is ready to :sheep:

It'd be nice to have the couple of todos captured as issues, I think you got one of them already.

jefflembeck:
I grabbed the promisifying of dropCache and of delSession and setSession. I _think_ that's everything. Merging into the feature branch.

PR #1778, opened 2016-01-21
absolute link path to nodejs.org

goenning:
Just a simple link fix


aredridel:
Whoa nice catch! Thank you!

PR #1785, opened 2016-01-21
Rework packages handler to be a simpler maze of logic

aredridel:


nexdrew:
Logic seems intact and is much easier to read now, LGTM! :+1:

PR #1786, opened 2016-01-21
Feature user restarts org

jefflembeck:
Well, here we go. This is the combined functionality of multiple issues and flows for restarting an organization.

We'll want to deploy this to staging, thoroughly check it, and also write newww-tests tests for it.

Once deployed to staging, I want people to walk through the functionality and the flows in this order:

@jonathanmarvens 
@ncawthon 
@stubbornella 
@seldo 
@ashleygwilliams 

Once everybody gives their sign off. We can put this into master.

Then, we dance.

:dancer:


aredridel:
:corkpop:

PR #1790, opened 2016-01-22
Redirect to payment-info upon restart

jefflembeck:
Instead of redirecting to the org front page with packages and what not,
restarting an org should take you to the payment-info page, where you
get to see action on what you've just done


stubbornella:
Where did this feature request come from? I'm not sure it makes sense in light of our story (coming up) to allow users to restart an org from the org page. I'm thinking they should be redirected to wherever they restarted it from, in either case with a success message.

jefflembeck:
@stubbornella it actually didn't come through, but the whole experience feels intensely awkward without doing it (in that we redirect to the org itself currently, so somebody can work on it - the first page feels weird)

jefflembeck:
This isn't necessary at this time. We have notifications now, etc.

PR #1794, opened 2016-01-23
Fix broken links in tutorial

wooorm:
I found a link in the tutorial on `https://www.npmjs.com/package/{package}/tutorial` which wasn’t clickable. I also changed the text inside the link from being a URL to a more helpful human readable text. If you’d rather like a different text I’m ok with changing it :)

issue

P.S. I found some weird mixture of spaces and tabs used for indentation which I didn’t change, but it might be worthwhile to look into.


ashleygwilliams:
LGTM!

wooorm:
:+1:

PR #1798, opened 2016-01-25
Bump npm-email-templates dependency and update npmo requirements URL

nexdrew:
Explicitly bump `npm-email-templates` to version `1.10.1` in order to pick up the latest [npmo welcome email](https://github.com/npm/email-templates/pull/21) (which is still not in staging or production for some reason).

Also update the URL we use for npmo installation requirements, which has its own page with updated content.

### DIFF
--- a/facets/enterprise/buy-license.js
+++ b/facets/enterprise/buy-license.js
@@ -159,7 +159,7 @@ module.exports = function(request, reply) {
             var sendEmail = request.server.methods.email.send;
 
             var data = {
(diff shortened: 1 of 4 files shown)


nexdrew:
This should probably be blocked by npm/email-templates#25 so we only need to explicitly bump the dependency once.

nexdrew:
No longer blocked, this should be good to go :+1:

aredridel:
LGTM!

nexdrew:
Verified new email in staging :sunglasses:

PR #1809, opened 2016-01-27
Add a shrinkwrap

aredridel:


aredridel:
I was thinking manually: with npm 3, we can update with --save, and the shrinkwrap is updated -- it just means we check in an exact pattern for how things should happen in prod, just as we've tested it. The workflow is relatively low friction that way, and doesn't need much automating.

aredridel:
Yeah, we do. Shrinkwraps really only work well on 3.

aredridel:
The shrinkwrap in this PR will have to be regenerated before merging, but yeah, let's get this outstanding and start the work.

aredridel:
@ceejbot @mmalecki any thoughts on how difficult it is to roll out npm 3? Any risks you can think of?

mmalecki:
@aredridel not hard at all. Also no risks I can think of right now.  I think @ceejbot had some concerns about initial npm 3 versions, but those problems should be solved by now.

aredridel:
Needs to be updated.

PR #1810, opened 2016-01-27
update catbox-redis request node-zendesk generic-pool google-libphonenumber

aredridel:


soldair:
looks good. 
I don't know the implications of the deps changes but the tests passed =)

aredridel:
Hehe. I mentioned the significant ones in the commit messages; mostly straightforward bumps. Some moves toward es6

test/adapters/bonbon.js
```diff
@@ -14,51 +14,47 @@ var Code = require('code'),
...
-    .get('/customer/bob/stripe').times(13)
-    .reply(200, {})
-    .get('/customer/mikeal/stripe')
-    .reply(200, {})
-    .get('/customer/seldo/stripe').times(4)
-    .reply(200, {});
-
-  done();
-});
-
```
rockbot: there's gotta be a better way to handle all of these things... the reason they were wrapped in the `before` block was so that there wouldn't be any mock leakage across test files. Putting them outside the `before` block (as you've done here) has the potential to do that :-/

test/adapters/redis-sessions.js
```diff
@@ -6,48 +6,23 @@ var Code = require('code'),
...
+  var prefix = "hapi-cache:%7Csessions:";
 
   it('creates a random hash for each user', function(done) {
-    bob1 = redisSessions.generateRandomUserHash('bob');
-    bob2 = redisSessions.generateRandomUserHash('bob');
-    alice1 = redisSessions.generateRandomUserHash('alice');
-
-    bobHash = redisSessions.userPrefixHash('bob');
-    aliceHash = redisSessions.userPrefixHash('alice');
-
```
rockbot: i like these changes a lot. wayyyyy cleaner - thank you!!

test/adapters/bonbon.js
aredridel: How so? It's whether or not the mocks had done() called on them that matters, right? Not the scope?

test/adapters/redis-sessions.js
aredridel: :)

test/adapters/bonbon.js
rockbot: Right, but if we forget to do `mock.done()`, then that's where things can go nastily awry.

aredridel: Isn't that true in either case? (the mock.done is still here, in the after block. All that changed was the scope.)

rockbot: I don't have the words to explain what I remember experiencing (aka doom) when it wasn't in the `before` block, but I'm cool with letting it ride and hopefully discovering that everything is still fine :-)

aredridel: Noted, and let's keep an eye out for it!

PR #1813, opened 2016-01-27
converts download model into an agent

rockbot:
very similar to https://github.com/npm/newww/pull/1371, but a tad cleaner and totally up to date with master :-)


aredridel:
LGTM!

PR #1815, opened 2016-01-28
User model to agent

rockbot:
we use the user agent a lot, hence lots of changes (but in the interest of not breaking anything between commits, this had to be one huge PR). apologies.

On the plus side: BYE BYE MODELS. WE HARDLY KNEW YE.

### DIFF
--- a/agents/user.js
+++ b/agents/user.js
@@ -8,39 +8,27 @@ var mailchimp = require('mailchimp-api');
 var P = require('bluebird');
 var Request = require('../lib/external-request');
 var userValidate = require('npm-user-validate');
+var USER_API = process.env.USER_API || 'https://user-api-example.com';
 var utils = require('../lib/utils');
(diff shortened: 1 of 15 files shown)


agents/user.js
```diff
@@ -396,14 +380,14 @@ User.prototype.signup = function(user, callback) {
     }, function() {
       // do nothing on success
     }, function(error) {
-      self.logger.error('Could not register user for npm Weekly: ' + user.email);
+      console.error('Could not register user for npm Weekly: ' + user.email);
```
aredridel: Ooh, let's use bole here rather than console.

rockbot: ooh, good call. i totally brain farted there :-P

aredridel:
Logging aside, this looks wonderful!

aredridel:
LGTM!

PR #1817, opened 2016-01-28
switch to new async npm-explicit-installs

bcoe:
This switches us to the new async version of `npm-explicit-installs`, this has some benefits:

* it now shows the actual publication dates/versions.
* it makes it easier to swap in a component fed by another data-source.

### DIFF
--- a/handlers/homepage.js
+++ b/handlers/homepage.js
@@ -2,11 +2,8 @@ var P = require('bluebird');
 
 var feature = require('../lib/feature-flags.js');
 var PackageAgent = require('../agents/package');
-<<<<<<< HEAD
 var DownloadAgent = require('../agents/download');
-=======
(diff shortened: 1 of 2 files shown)


handlers/homepage.js
```diff
@@ -2,7 +2,11 @@ var P = require('bluebird');
 
 var feature = require('../lib/feature-flags.js');
 var PackageAgent = require('../agents/package');
+<<<<<<< HEAD
 var DownloadAgent = require('../agents/download');
+=======
+var ExplicitInstalls = require("npm-explicit-installs")
+>>>>>>> update to async/up-to-date version of npm-explicit-installs
```
rockbot: ... I don't think this is what you want ;-)

aredridel:
Looks great! Ship it if the tests pass!

PR #1820, opened 2016-01-29
removes download service, which hasn't been used for, oh, well, _ages_

rockbot:
bye bye old code!


aredridel:
LGTM And good riddance!

PR #1823, opened 2016-02-01
Fix integration tests

rockbot:
There are a few flaky tests, which seems to have been due to funky race conditions. These have been fixed, thanks to the beauty of promises!

Also, there may have been some ES6. Maybe. ;-)

cc: @aredridel


jefflembeck:
:sheep: \o/

PR #1824, opened 2016-02-02
turns email service into an adapter

rockbot:
moving forward, we're going to leave plugins for hapi-specific code, and allow our own code to run as separate modules. This will allow us to slowly move our codebase into a more modular direction.

this is the first step as we say goodbye to services.


aredridel:
This is delightful!

PR #1825, opened 2016-02-02
some more code changes for integration tests

rockbot:
once again, dear readers, we have flaky integration tests, almost entirely due to race conditions. this is my attempt to remove as many of said race conditions as I can find, though i'm sure there are more... lurking in the shadows... harrumph.


aredridel:
Yesss!

PR #1826, opened 2016-02-02
removes email service from plugins

rockbot:
totally forgot to remove the email service from the plugins file (which is not tested via unit tests - another good thing about this transition away from non-hapi-based plugins).

whoops :-(


aredridel:
Lgtm!

PR #1828, opened 2016-02-03
First take at some marketing analytics.

jonathanmarvens:
Added analytics tracking for the promotions.


jonathanmarvens:
__CC:__ @npm/dubdub, @cowperthwait.

cowperthwait:
@jonathanmarvens I don’t actually understand how this works, so I trust you. At some point when we’re in mutually convenient time zones, want to do a quick 5-min Slack to talk me thru the Google Analytics parts?

adapters/bonbon.js
```diff
@@ -89,6 +90,10 @@ exports.register = function(server, options, next) {
       env: {
         NPMO_COBRAND: process.env.NPMO_COBRAND,
         CANONICAL_HOST: process.env.CANONICAL_HOST
+      },
+      ga: {
+        dimensions: JSON.stringify(qs.parse(process.env.PROMOTION_GA_DIMENSIONS)),
+        id: process.env.GA_ID
```
aredridel: A fallback to the old ID with `|| 'UA-47041310-1'` here would make this deployable with no new config active.

templates/layouts/default.hbs
```diff
@@ -62,13 +62,15 @@
   <script src="/static/{{m "js/index.min.js"}}"></script>
   {{/if}}
 
+  <meta name="x-npm:ga:id" content="{{{ga.id}}}">
```
aredridel: Yeah, this can be just `{{` not `{{{` -- if there were any escaping to do, we'd want to do it.

aredridel:
This looks pretty good. A couple comments, and Raquel's questions stand!

jonathanmarvens:
@rockbot:

1. The content is coming from our CMS API, but some of the `{{{` usage here was accidental and not what we want, so thanks for catching it!
2. Re `PROMOTION_GA_DIMENSIONS`:
  - Just a query string: `'aboveChrome=dimension1&belowHeader=dimension3&packagePage=dimension2'`.
  - AFAIK, it's fine if anyone sees it.

templates/layouts/default.hbs
jonathanmarvens: @aredridel: Fixed. Thanks for catching that!

adapters/bonbon.js
jonathanmarvens: @aredridel: IDK how I feel about that, TBH. If we're gonna go with using an environment variable for this, is the fallback necessary?

aredridel: Nope! It'll change deployment strategy -- what we have to do to launch this -- but I'm happy with whichever way you decide. What happens if we screw up deployment and get the variable name or value wrong?

aredridel:
Agreed, those values are not private or secret.

templates/layouts/default.hbs
aredridel: \o/

adapters/bonbon.js
jonathanmarvens: @aredridel: I understand the reasoning, but I'd prefer to leave it as, unless there's a strong case for going that route. OK?

jonathanmarvens:
@npm/dubdub: OK to merge?

aredridel:
\o/

PR #1829, opened 2016-02-03
upgrade to new version of npm-explicit-installs

bcoe:
I've upgraded to the newest version of `npm-explicit-installs`, this version:

* supports proxies (which is important for big companies).
* has a bin for editing the packages on the home-page (this is only important for `npmo`).
* I also fixed a display bug I noticed: npmo has no user pages, but we still linked to user-pages.


aredridel:
LGTM!

PR #1833, opened 2016-02-04
Fix inconsistent use of ' and "

briandela:
There were some places where " was being used but looks like ' is used everywhere else.


aredridel:
Thank you, but this isn't actually something we want to do. We handle low-level style stuff like this with  `esformatter`, and it doesn't have a rule for this set up.

PR #1835, opened 2016-02-04
Step 1/n for moving npme service to customer agent

rockbot:
lest I create a PR that's far too big for a proper review (and lest I find myself in a weird anti-pattern), I'm breaking my commits up into individual PRs.

Of particular note in this PR: the customer agent originally had a requirement that a username had to be passed into the constructor. To move the npme service over, we need to use some of the functions _without_ a name. (Particularly since the on-site flow currently does not need a user to log in. This will change eventually.)

Thoughts? Recommendations? Reservations?


agents/customer.js
```diff
@@ -1,28 +1,31 @@
-var _ = require('lodash');
-var assert = require('assert');
+// var _ = require('lodash');
```
aredridel: Can just rip these lines out!

aredridel:
LGTM. I see no reason to assert there, it's not a pattern we're generally using.

aredridel:
:sheep:

PR #1839, opened 2016-02-05
Fix package test

aredridel:


aredridel:
Well this got fixed independently so nevermind.

PR #1840, opened 2016-02-05
Adding rejection obj to catch promise-based errors

aredridel:
These changes came about because there was a failing test, but it wasn't failing when we ran the tests, despite a "failing" print out.

This happens when a test fails inside of a catch block.

Let's make sure this doesn't happen again. Now tests that fail in the catch blocks will stop the test runner

The problem with change three is that it's a global change. We should talk about where that goes.

### DIFF
--- a/test/agents/team.js
+++ b/test/agents/team.js
@@ -3,13 +3,36 @@ var Code = require('code'),
   lab = exports.lab = Lab.script(),
   describe = lab.experiment,
   it = lab.test,
+  afterEach = lab.afterEach,
   expect = Code.expect,
   nock = require('nock'),
   fixtures = require('../fixtures');
 
 var Team = require('../../agents/team');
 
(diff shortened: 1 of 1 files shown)


aredridel:
Yeah, this is a global thing that has to go _somewhere_ if we want it. It's kind of a thing a test framework like lab should do -- perhaps upstream there! -- but it's worth considering.

jonathanmarvens:
I honestly feel that __newww__ is the wrong place to solve this particular problem.

aredridel:
What would be an improvement?

jonathanmarvens:
@aredridel: Contribute a fix upstream instead?

aredridel:
Possibly, how do we get this working in the mean time?

aredridel:
(I think that's legit -- but it'd be nice to vet the idea in place and know it works)

PR #1843, opened 2016-02-05
removes getCustomer from npme service (duplicate of customer agent's getById)

rockbot:
The error handling in this PR isn't my favorite, as it seems there are some relics from the days of "a 404 is an error" misunderstanding. My gut says it's more important to move stuff out of the service than to fix all the things right this second, so I'd like to fix it in some future work (aka fixing the on-site purchase flow).

Another bit of future work: promisify _everything_. The callbacks are rampant in this entire flow!


jonathanmarvens:
LGTM, @rockbot!

PR #1849, opened 2016-02-08
Added a blacklist for packages.

jonathanmarvens:
This is just one possible way to handle this.


aredridel:
Hmm. Word blacklists have never struck my fancy. Always incomplete. False positives possible too.

jonathanmarvens:
Same here, @aredridel, but now, we can have a discussion here on why this isn't the best option!

aredridel:
Is there a reason we went for a word blacklist? Why not just block the issac/9999/blacklist packages, making it in control of the registry? (Ultimately, there's supposed to be a 'tombstone' sort of package that prevents republication)

I think a reactive, not proactive approach is actually the right one (And historically what we've done)

stubbornella:
We'll be moving over to some feature work for a bit, we can reopen this discussion when we get back to it.

PR #1850, opened 2016-02-08
Updated "truncate" package dependency.

jonathanmarvens:
There was a breaking change, but that's taken care of.


aredridel:
Ooh nice! Proper ellipsis! :sheep: it!

PR #1851, opened 2016-02-08
Updated "cheerio" package dependency.

jonathanmarvens:


aredridel:
Is 0.19 to 0.20 a breaking change?

jonathanmarvens:
Argh. I don't remember exactly, TBH, @aredridel, but if it was, it definitely wasn't something that should affect us. (I should've kept notes!)

jonathanmarvens:
@aredridel: https://github.com/cheeriojs/cheerio/blob/f500197155d4e03e7e03fa51925d9cc0aca778fc/History.md ← Here are the release notes.

aredridel:
Looks shippable. I wish they'd adopt unambiguous semver.

PR #1855, opened 2016-02-08
Use per-version cache keys for package readmes and descriptions

aredridel:
This will make new versions appear instantly rather than waiting for cache timeout.


jonathanmarvens:
LGTM, @aredridel.

aredridel:
Thank you @jonathanmarvens !

PR #1863, opened 2016-02-09
Add new relic fe

jefflembeck:
This PR will allow us to handle performance monitoring with New Relic for both server and browser-side monitoring.

The major caveat at this point: the change to the csp allows unsafe-inline. This could be incredibly problematic.

I would like @evilpacket to take a look at this to see if there is a way we can make this work appropriately.

Environment variables that must be set for this to work:

```
NEW_RELIC_APP_NAME
NEW_RELIC_LICENSE_KEY
NEW_RELIC_LOG_LEVEL
NEW_RELIC_NO_CONFIG_FILE
```

### DIFF
--- a/lib/csp.js
+++ b/lib/csp.js
@@ -4,6 +4,7 @@ csp.default = {
   defaultSrc: 'self',
   scriptSrc: [
     'self',
+    'unsafe-inline',
 
     // Hubspot
     'https://forms.hubspot.com/uploads/form/v2/419727/9a2b4ac5-ef09-43e6-854a-d82c92347c9d',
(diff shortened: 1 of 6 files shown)


lib/csp.js
```diff
@@ -4,6 +4,7 @@ csp.default = {
   defaultSrc: 'self',
   scriptSrc: [
     'self',
+    'unsafe-inline',
```
aredridel: Hmm. can we use the checksummed inline script features instead of unsafe-inline?

evilpacket: It would be nice to use a nonce here instead of unsafe-inline however blankie doesn't quite yet support nonces. I've asked @nlf if he could add support for that and it's in his queue, but he's afk this week.

aredridel:
What are the deployment constraints here?

jefflembeck:
@aredridel the thing that's holding us back in this case is our CSP. I'd like to make the change necessary for that to occur -- i'll message you about that. After that, it's setting ENV vars and out it can go.

aredridel:
Yeah, the CSP is the interesting case!

PR #1867, opened 2016-02-09
createTrial: from npme service to customer agent

rockbot:
slowly but surely, we'll get there!


aredridel:
nice!

agents/customer.js
```diff
@@ -127,6 +127,75 @@ Customer.prototype.createOnSiteLicense = function(licenseDetails, callback) {
...
+
+        if (resp.statusCode >= 400) {
+          err = new Error(newTrial);
+          err.statusCode = resp.statusCode;
+          return reject(err);
+        }
+
+        return accept(newTrial);
+      });
+    }
```
jonathanmarvens: @rockbot: Why not pull this function (`createNewTrial(...)`) out on its own?

rockbot: no good reason :-)

jonathanmarvens: @rockbot: I personally would extract it and then maybe add a test or two for it? It's alright as is though.

jonathanmarvens:
@rockbot: :+1:.

Issue #1039, opened 2015-06-13
Auto-focus the searchbar?


sinkingshriek:
I'm being really picky here and I'm aware that there's a way to do this which is mentioned [here](https://github.com/npm/newww/issues/513) but for those who don't know this trick it'd be really helpful if the search was auto-focused just like how google does it because the number 1 reason for me to come to npmjs.com is to search for packages. 
It's also a really simple change which just needs `autofocus` attribute [here](https://github.com/npm/newww/blob/master/templates/layouts/default.hbs#L96) so sending a PR just for one word sounded a little ridiculous to me. Let me know how you guys feel about this. 
Again, I'm being extremely picky but it's nice to have for new comers.

isaacs:
Roughly 50% of users believe that the search bar should be auto-focused, and don't see how anyone could ever want it otherwise.  The other 50% feel about as strongly the opposite.  This option has flipped around a lot over the years.

Maybe it could be a user preference?

sinkingshriek:
Hmm. I wonder why people wouldn't want that to be auto focused. Maybe because it doesn't let you use your arrow key to scroll down? Or perhaps in mobile it brings up keyboard? 
I would like to know what the use cases are for the other 50% of people who thinks it's not a good idea.
Regardless, having that in user preference seems kind of overkill but I suppose if others make a really good case for it then I wouldn't be opposed of that idea.

aredridel:
Autofocus destroys accessibility by text to speech, so it cannot be on by default.

aredridel:
And as a user preference, this does in fact seem like overkill, so I'm going to close this issue.

aredridel:
Also: press `/` on the page and the box will be focused.

sinkingshriek:
Fair enough. I knew about the `/` but only after digging into source code I believe. If npm site does end up having lot of shortcuts in future then document like [github](https://help.github.com/articles/using-keyboard-shortcuts/)'s would be tremendous help for new comers. But atm, I'm okay with this.

aredridel:
Indeed! making that discoverable would be good to think about!

PR #1869, opened 2016-02-10
Small UX Improvment.

bhaveshgohel:
Add Autofocus attribute in "find packaegs" search input field. So on page load we can search right away of packages.


KenanY:
https://github.com/npm/newww/issues/1039?

bhaveshgohel:
@KenanY : Sorry I was not aware about it... But i strongly feel that this is a great for UX in my opinion.

aredridel:
Yeah, this is actually a net usability loss for many people; it confuses accessibility tools, and makes the pages hard to navigate, and prioritizes the search use case (which is needed!) over others (which are also needed!)

However, there is the `/` hotkey to jump to the search bar, which should give you the user affordances you're looking for.

PR #1871, opened 2016-02-10
Hid blacklisted packages on @isaacs profile page.

jonathanmarvens:


handlers/packages.js
```diff
@@ -7,11 +8,10 @@ module.exports = function(request, reply) {
   new User(request.loggedInUser)
     .getPackages(name, offset)
     .then(function(packages) {
-      return reply(packages).code(200);
+      reply((name === 'isaacs') ? filterBlacklistedPackages(packages) : packages).code(200);
```
rockbot: why no `return` here?

handlers/packages.js
```diff
@@ -7,11 +8,10 @@ module.exports = function(request, reply) {
...
     })
     .catch(function(err) {
       request.logger.error(err);
-      return reply('error getting packages').code(500);
+      reply('error getting packages').code(500);
```
rockbot: also here? :)

handlers/packages.js
jonathanmarvens: @rockbot: It's not needed.

handlers/packages.js
jonathanmarvens: @rockbot: No need for it here either!

aredridel:
Tidy! Works for me!

PR #1880, opened 2016-02-11
Make those new SPDX license links actually work with r-r data

aredridel:


aredridel:
Got some :eyes: from @kemitchell and we're going with it!

evilpacket:
:lock:

PR #1887, opened 2016-02-15
Just use an image rather than iconic SVG inlining for actions button

### LINKED ISSUE #1806: "Delete this member" button is missing an icon
"Actions" alt text is visible where there should be a button.

screenshot 2016-01-27 11 26 20

aredridel:
Fixes #1806


aredridel:
Jeff says +1, so I'ma go this!

PR #1889, opened 2016-02-16
Break long package names

fredericmarx:
Break long package names so they won't extend past the viewport.

Example: https://www.npmjs.com/package/ifyouwanttogetthesumoftwonumberswherethosetwonumbersarechosenbyfindingthelargestoftwooutofthreenumbersandsquaringthemwhichismultiplyingthembyitselfthenyoushouldinputthreenumbersintothisfunctionanditwilldothatforyou

### DIFF
--- a/assets/styles/typography.styl
+++ b/assets/styles/typography.styl
@@ -182,6 +182,7 @@ pre
 &.package-name
   padding-bottom 0
   padding-right 25px
+  word-break: break-word;
   a
     color text-color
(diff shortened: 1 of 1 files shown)


wesbos:
Probably want to remove that colon too - they are using stylus as bare as it gets!

fredericmarx:
Ha! Could have figured that out myself. Thanks for the heads up!

PR #1890, opened 2016-02-17
Make "log out" fully visible

leo:
I'm not sure, if it's just a issue in Safari, but:

**BEFORE**

screen shot 2016-02-17 at 18 51 41

**AFTER**

screen shot 2016-02-17 at 18 52 38


leo:
Design of menu changed.

PR #1903, opened 2016-03-06
Allow for keywords to be links, just like in other keyword lists

krainboltgreene:
See: https://github.com/npm/newww/blob/master/templates/helpers/packageKeywords.js#L10-L15

Not quite sure on the hbs syntax, honestly.


krainboltgreene:
Also, while I added title, there may be other accessibility details I'm missing.

PR #1910, opened 2016-03-14
Add ember to what npm is for

Robdel12:
Noticed it wasn't there. Figured I'd add it :P


Robdel12:
Closing since the front page no longer has this

PR #1915, opened 2016-03-23
get a job => get hired

balupton:
At least in Australia, "get a job" is quite insulting... it is only ever used to imply someone is a homeless bum that is draining the economy... especially as the presupposition of it is that you don't have a job, so such connotations are reasonably expected... "get hired" is a good alternative that doesn't imply such a thing, as the presupposition is just you want to be hired, rather than you are missing anything.

### DIFF
--- a/templates/homepage.hbs
+++ b/templates/homepage.hbs
@@ -75,7 +75,7 @@
 
   {{#unless features.npmo}}
     <li>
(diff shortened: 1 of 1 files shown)


balupton:
Seems something else broke the build...

aredridel:
Thanks!

Issue #1933, opened 2016-05-07
Table not respecting markdown cell alignment.


jdalton:
Tables are not respecting markdown cell alignment.
See [package/lodash-webpack-plugin](https://www.npmjs.com/package/lodash-webpack-plugin#feature-sets) for an example of the issue.

The cells are aligned with `:---` which signals left alignment.

```md
| Feature | Description  |
|:---|:---|
| `shorthands`   | Iteratee shorthands for `_.property`, `_.matches`, & `_.matchesProperty`. |
| `collections`  | Support for objects in “Collection” methods like `_.each`, `_.filter`, & `_.map`. |
| `currying`     | Support for `_.curry` & `_.curryRight`. |
```
which should render as

| Feature | Description  |
|:---|:---|
| `shorthands`   | Iteratee shorthands for `_.property`, `_.matches`, & `_.matchesProperty`. |
| `collections`  | Support for objects in “Collection” methods like `_.each`, `_.filter`, & `_.map`. |
| `currying`     | Support for `_.curry` & `_.curryRight`. |

vitali-ausianik:
Did someone have a chance to take a look on this issue?

revin:
The sanitizer used in [npm/marky-markdown](https://github.com/npm/marky-markdown) is stripping the `style` attributes from the rendered `<th>` and `<td>` elements. @jdalton would you terribly mind moving this over there for us to talk about?

revin:
actually @jdalton never mind I got it covered; opened [npm/marky-markdown#212](https://github.com/npm/marky-markdown/issues/212) for this.

jdalton:
Moved to https://github.com/npm/marky-markdown/issues/212.

Issue #1793, opened 2016-01-23
Markdown CSS Rendering: Table Content Is Centered By Default


jhuckaby:
Table cells should align left unless explicitly set otherwise in the Markdown source. On npmjs.com they are all centered by default:

<img width="750" alt="screenshot 2016-01-22 19 59 32" src="https://cloud.githubusercontent.com/assets/700765/12528141/ac14b22e-c142-11e5-9624-628ae302d804.png">

You can see the issue live here:
https://www.npmjs.com/package/npm-markdown-bugs#tables-centered-by-default

Compare to the same Markdown on GitHub here:
https://github.com/jhuckaby/npm-markdown-bugs#tables-centered-by-default

This was originally posted in the marly-markdown repo issue tracker here: npm/marky-markdown#126 -- but was determined that it was probably a CSS issue on the npmjs.com site instead.

- Joe

vitali-ausianik:
Guys, please move it up in priority, this is really uncomfortable issue

PR #1943, opened 2016-05-19
Set left alignment by default for table td - #1933, #1793

vitali-ausianik:
Issues #1933, #1793 
Set "left" alignment by default for table td
Set "center" alignment by default for table th

For now npmjs.com renders tables with "center" alignment by default. It makes readme' of many packages a bit unreadable, because github uses "left" alignment for tables columns by default and "center" alignment for table' headline (th)


vitali-ausianik:
Looks like tests is a bit broken - according to travis logs it cannot even install dependencies :(

PR #1964, opened 2016-07-17
Revert "Use handlebars {{ instead of {{{ to escape html on package description"

### LINKED ISSUE #1789: README ampersands are encoded
moving from: https://github.com/npm/marky-markdown/issues/124 cc/ @lewiscowper @revin

099e6090-c10a-11e5-8587-c67a956d6c20

addaleax:
Reverts #1731 to fix #1789

Reasoning is pretty much [here](https://github.com/npm/newww/issues/1789#issuecomment-223802559). TL;DR: That change broke about 1 % of package descriptions *and* didn’t even make the package description which was supposed to be fixed by #1731 render correctly (i.e. as intended by the author).

### DIFF
--- a/templates/package/access.hbs
+++ b/templates/package/access.hbs
@@ -2,7 +2,7 @@
   <h1 class="package-name centered">
     <a href="/package/{{name}}">{{name}}</a>
   </h1>
-  <p class="package-description centered">{{description}}</p>
+  <p class="package-description centered">{{{description}}}</p>
 {{/with}}
 
 {{> package-header package}}
--- a/templates/package/show.hbs
+++ b/templates/package/show.hbs
@@ -10,7 +10,7 @@
       <i class="icon-public"></i>
     {{/if}}
   </h1>
-  <p class="package-description">{{description}}</p>
+  <p class="package-description">{{{description}}}</p>
 
   <form class="star">
     <input type="hidden" name="name" value="{{name}}">


aredridel:
There are security implications to this.

addaleax:
The content is parsed as markdown beforehand, though, so I don’t think it’s different from e.g. the displayed README content? I’ve also linked it in the issue, but [here](https://github.com/npm/newww/blob/db1f3dbb183f43c886da6eb25ae214370649c1e2/presenters/package.js#L211-L234) is the parsing step for the package description.

That’s why the double escaping described in #1789 comes up in the first place.

addaleax:
I.e. the markdown parser already takes care of sanitizing any HTML contained in the markdown, that’s why not escaping again by using `{{{` here was okay to begin with.

aredridel:
Turns out this field only sometimes comes from markdown. It's a combination of two data sources with different needs there. It's _tricky_

addaleax:
If you’re talking about the cache, that should be fine, the data is already stored there in parsed form.

But yeah, you probably know your way around this codebase better than I do, and if you say there’s more to be done about this, I believe you. I’d love to see this fixed, though, and I’m happy to do the work.

aredridel:
Yeah. It actually comes from both the `description` and the first line of markdown -- the interface there has to be made consistent (to always do escaping at the lower level) -- then at the higher layer use `{{{` again. It's got a couple layers of dependencies but I think the actual next change may have to be in marky-markdown. I'd have to trace the task dependencies to find it again.

addaleax:
fwiw, I just “accidentally” published a new package version without a `description:` key in the package.json, and it shows the same doubly-escaped behaviour: https://www.npmjs.com/package/checkref

aredridel:
Yeah. It does some fuzzy matching to decide whether to display the h1 from the markdown or the description and it's ... kinda terrible!

